PR #2, opened 2016-09-07
Do not run test_read_attribute on appveyor

enzbang:
The test is not working and it is difficult to reproduce this outside
appveyor environment.

Change-Id: I530620c9220e8ac1f56cc0239cc292e5f0229ec0


enzbang:
Merged!

Issue #15, opened 2016-11-11
use isort to automatically add the required imports


enzbang:
I’m suggesting to use [isort](https://github.com/timothycrosley/isort) so that we can automatically add the required `from future` imports.

The required config would be:

`[isort]`
`balanced_wrapping = True`
`add_imports =`
`    from __future__ import absolute_import`
`    from __future__ import division`

And maybe:
`    from __future__ import print_function`
see #14

Issue #14, opened 2016-11-11
Use from __future__ import division


enzbang:
In order to reduce the gap between Python 2 and Python 3 I would suggest to use:

`from __future__ import division`

This should not have a big impact on the code, the initial patch is pretty small.

PR #16, opened 2016-11-12
Import division from future and sort all imports

enzbang:
Use isort to automatically add:

`    from __future__ import absolute_import`
`    from __future__ import division`
`    from __future__ import print_function`

in each e3 modules.

See #14 and #15


Nikokrock:
Looks good

Nikokrock:
looks good

PR #17, opened 2016-11-14
increase test coverage e3.fs to 76% and e3.net.smtp to 93%

piratos:
**tests e3.fs**
- added hidden files and unicode tests
- added copy symlink test

**tests e3.net.smtp**
- added different exceptions return from the mocked SMTP object
- added max size test
- added mail sent successfully scenario

**e3.net.smtp**
- added #all: no coverage of the system sendmail binaries

### DIFF
--- a/tests/tests_e3/fs/main_test.py
+++ b/tests/tests_e3/fs/main_test.py
@@ -54,6 +54,14 @@ def test_cp():
     assert os.path.exists(os.path.join(dest3, 'a1'))
 
 
+@pytest.mark.skipif(sys.platform == 'win32', reason='test using symlink')
+def test_cp_symplink():
+    e3.os.fs.touch('c')
+    os.symlink('c', 'c_sym')
+    e3.fs.cp('c_sym', 'd', preserve_symlinks=True)
+    assert os.path.islink('d')
+
+
 def test_echo():
     dest_file = 'echo_test'
     e3.fs.echo_to_file(dest_file, 'foo')
@@ -107,6 +115,15 @@ def test_tree_state():
     e3.os.fs.touch('toto')
     state4 = e3.fs.get_filetree_state(current_dir)
     assert state4 != state3
+    hidden = os.path.join(current_dir, '.h')
+    e3.fs.mkdir(hidden)
+    state5 = e3.fs.get_filetree_state(current_dir)
+    assert state5 == state4
+    e3.os.fs.touch('.toto')
+    state6 = e3.fs.get_filetree_state(current_dir)
+    assert state6 == state5
+    state6 = e3.fs.get_filetree_state('toto')
+    assert isinstance(state6, str)
 
 
 @pytest.mark.skipif(sys.platform == 'win32', reason='test using symlink')
--- a/tests/tests_e3/net/smtp/main_test.py
+++ b/tests/tests_e3/net/smtp/main_test.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 import mock
 import e3.net.smtp
-
+import smtplib
 
 def test_sendmail():
     from_addr = 'e3@example.net'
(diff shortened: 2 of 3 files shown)


tests/tests_e3/fs/main_test.py
```diff
@@ -138,6 +153,7 @@ def test_sync_tree_with_symlinks():
         b,
         os.path.join(m2, 'c')) == ''
     assert e3.diff.diff(
+
```
enzbang: why an empty line here?

tests/tests_e3/net/smtp/main_test.py
```diff
@@ -21,3 +21,46 @@ def test_sendmail():
...
+
+    with mock.patch('smtplib.SMTP') as mock_smtp:
+        mock_smtp.side_effect = smtplib.SMTPException()
+        result = e3.net.smtp.sendmail(
+            from_addr,
+            to_addresses,
+            msg_as_string,
+            ['smtp.localhost'])
+        assert result == False
+
```
enzbang: You should never have two empty lines in an indented block. But apparently we’re not running any style checks on our tests. I think I’ll add it.

tests/tests_e3/net/smtp/main_test.py
```diff
@@ -21,3 +21,46 @@ def test_sendmail():
             from_addr,
             to_addresses,
             msg_as_string)
+
+def test_sendmail_onerror():
+    from_addr = 'e3@example.net'
+    to_addresses = ['info@example.net', 'info@example.com']
+    msg_as_string = 'test mail content'
+    msg_size_excceed = "A"*1200
```
enzbang: I guess msg_size_exceed (typo)

tests/tests_e3/net/smtp/main_test.py
```diff
@@ -21,3 +21,46 @@ def test_sendmail():
...
+    result = e3.net.smtp.sendmail(
+        from_addr,
+        to_addresses,
+        msg_size_excceed,
+        ['smtp.localhost'],
+        max_size=8/1024)
+    assert result == False
```
enzbang: when comparing against False/True always use *is*. 

So here assert `result is False`

tests/tests_e3/net/smtp/main_test.py
```diff
@@ -21,3 +21,46 @@ def test_sendmail():
...
+
+    with mock.patch('smtplib.SMTP') as mock_smtp:
+        mock_smtp.side_effect = smtplib.SMTPException()
+        result = e3.net.smtp.sendmail(
+            from_addr,
+            to_addresses,
+            msg_as_string,
+            ['smtp.localhost'])
+        assert result == False
```
enzbang: same here `assert result is False`

tests/tests_e3/net/smtp/main_test.py
```diff
@@ -21,3 +21,46 @@ def test_sendmail():
...
+
+    with mock.patch('smtplib.SMTP') as mock_smtp:
+        smtp_mock = mock_smtp.return_value
+        smtp_mock.sendmail.side_effect = smtplib.SMTPException()
+        result = e3.net.smtp.sendmail(
+            from_addr,
+            to_addresses,
+            msg_as_string,
+            ['smtp.localhost'])
+        assert result == False
```
enzbang: here also ;)

tests/tests_e3/net/smtp/main_test.py
```diff
@@ -21,3 +21,46 @@ def test_sendmail():
...
+
+    with mock.patch('smtplib.SMTP') as mock_smtp:
+        smtp_mock = mock_smtp.return_value
+        smtp_mock.sendmail.return_value = {}
+        result = e3.net.smtp.sendmail(
+            from_addr,
+            to_addresses,
+            msg_as_string,
+            ['smtp.localhost'])
+        assert result == True
```
enzbang: assert `result is True` is that case

and add the missing newline and the end of the file

PR #21, opened 2016-11-15
Review implementation of wait_for_processe on Unix

Nikokrock:
Use an implementation based on pipe and signal handler. Previous implementation can present race conditions as waitid can always present the same process if no wait is done on it


e3/os/process.py
```diff
@@ -625,29 +625,49 @@ def wait_for_processes(process_list, timeout):
...
+                # Do a first check in case a SIGCHLD was emited before the
+                # initialisation of the handler.
+                for index, p in enumerate(process_list):
+                    if p.poll() is not None:
+                        return index
+
+                # Wait for a sigchld signal. Note that select might
+                # be interrupted by signals thus the loop
+                while True:
+                    select_args = [[fd_r], [], []]
```
enzbang: do we need to keep select_args in the loop? can we move that before `while remain >=...`?

Issue #27, opened 2016-11-26
appveyor build broken


enzbang:
While trying to understand why my pull-request was not working on appveyor I decided to force a rebuild of the master branch. And it appears that test_deploy_sandbox is now failing with:

```DEBUG    Run: cd c:\users\appveyor\appdata\local\temp\1\tmpfep49d; 'bin\anod' download a
ERROR    [Error 2] The system cannot find the file specified
ERROR    bin\anod not found
```

enzbang:
See https://github.com/pypa/setuptools/issues/861

That was a bug in setuptools, now fixed.

PR #28, opened 2016-11-26
Work around bug in setuptools

enzbang:
issue #27


enzbang:
Setuptools bug has been fixed.

PR #30, opened 2016-11-30
Increase coverage of e3.os.fs to 83%

piratos:
tests added for

- df

- ls

- max_path

### DIFF
--- a/tests/tests_e3/os/fs/main_test.py
+++ b/tests/tests_e3/os/fs/main_test.py
@@ -7,6 +7,8 @@ import tempfile
 import e3.fs
 import e3.os.fs
 import pytest
+import mock
+import numbers
 
 
 @pytest.mark.xfail(sys.platform == 'win32',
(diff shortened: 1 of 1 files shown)


tests/tests_e3/os/fs/main_test.py
```diff
@@ -55,3 +57,33 @@ def test_mkdir_exists(caplog):
...
+    e3.os.fs.touch("toto")
+    with mock.patch("os.link") as mock_link, mock.patch("shutil.copy2") as mock_copy2:
+        e3.os.fs.ln("toto", "tata")
+        assert any([mock_link.called, mock_copy2.called])
+    e3.os.fs.touch("tata")
+    os.chmod("tata", 0000)
+    try:
+        e3.os.fs.ln("toto", "tata")
+        assert False
+    except:
```
Nikokrock: except Exception here

PR #31, opened 2016-12-05
remove external binaries usage from e3.archive

piratos:
remove the usage of the external binaries in both

- e3.archive.unpack_archive
and
- e3.archive.create_archive

### DIFF
--- a/e3/archive.py
+++ b/e3/archive.py
@@ -3,9 +3,10 @@ from __future__ import absolute_import, division, print_function
 
 import fnmatch
 import os
-import subprocess
 import sys
+import tarfile
 import tempfile
+import zipfile
 from contextlib import closing
 
 import e3
@@ -13,7 +14,6 @@ import e3.error
 import e3.fs
 import e3.log
 import e3.os.fs
-from e3.os.process import Run
 
 logger = e3.log.getLogger('archive')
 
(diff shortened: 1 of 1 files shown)


e3/archive.py
```diff
@@ -44,12 +44,14 @@ def __select_archiving_tool(filename, unpack=True,
     :return: a tuple (ext, True if we should use python libraries else False)
     :rtype: (str, bool)
     """
+
     def has_python_lib_support(archive_ext):
```
Nikokrock: No need anymore for has_python_lib. We assume that we have a full python and thus tarfile and zipfile are always there

e3/archive.py
```diff
@@ -58,69 +60,42 @@ def has_python_lib_support(archive_ext):
...
 
-    if sys.platform == 'win32' or require_wilcard:
-        # On windows, do not spawn tar/zip often provided by cygwin but calls
-        # tarfile/zipfile python implementation.  If wildcards (* or ?) are
-        # used in selected_files, to avoid portability issue, use directly the
-        # python library if possible
-        impls = (has_python_lib_support, has_binary_tools)
+    if has_python_lib_support(ext):
+        return ext, True
     else:
```
Nikokrock: No need for that (for the same reason as before)

e3/archive.py
```diff
@@ -424,49 +374,6 @@ def create_archive(filename, from_dir, dest, tar='tar', force_extension=None,
             archive.add(from_dir, from_dir_rename, recursive=True)
             archive.close()
     else:
```
Nikokrock: Remove that has we always assume that tarfile and zipfil are present

PR #52, opened 2017-02-10
Add support Ctrl-C interruption in e3.job.scheduler

Nikokrock:
* add recursive parameter to Run.kill.
* when interrupt is called on a job mark set attribute interrupted
  to True.
* for ProcessJob ensure that proc_handle is set before the end of
  the process to allow interrupt and in case the job is interrupted
  perform a recursive kill.
* in Scheduler add parameter job_timeout that define the maximum amount
  of time a job can run.
* in Scheduler handle Ctrl-C nicely (try to kill all active jobs and
  call collect for them before reraising the exception)

### DIFF
--- a/e3/job/scheduler.py
+++ b/e3/job/scheduler.py
@@ -2,12 +2,12 @@ from __future__ import absolute_import, division, print_function
 
 from collections import deque
 from datetime import datetime
-from Queue import Queue
+from Queue import Empty, Queue
 
 import e3.log
 from e3.collection.dag import DAGIterator
 
-logger = e3.log.getLogger('e3.job.scheduler')
+logger = e3.log.getLogger('job.scheduler')
 
 
 class Scheduler(object):
@@ -17,7 +17,8 @@ class Scheduler(object):
                  job_provider,
                  collect=None,
                  queues=None,
-                 tokens=1):
+                 tokens=1,
+                 job_timeout=3600 * 24):
         """Initialize Scheduler.
 
         :param job_provider: function that returns instances of Job.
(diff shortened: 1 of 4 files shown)


e3/job/scheduler.py
```diff
@@ -127,13 +134,20 @@ def run(self, dag):
...
+        try:
+            while not self.is_finished:
+                self.enqueue()
+                self.launch()
+                self.log_state()
+                self.max_active_jobs = max(self.max_active_jobs,
+                                           len(self.active_jobs))
+                self.wait()
+        except KeyboardInterrupt:
+            logger.info('Ctrl-C received')
```
enzbang: you should add “interrupting jobs...”

e3/job/scheduler.py
```diff
@@ -183,7 +197,13 @@ def wait(self):
             return
 
         # Wait for message from one the active jobs
-        uid = self.message_queue.get(True, None)
+        while True:
```
enzbang: please add more comments here

e3/job/scheduler.py
```diff
@@ -183,7 +197,24 @@ def wait(self):
...
+            try:
+                # The first job in active jobs is the oldest one
+                # compute the get timeout based on its startup information
+                deadline = datetime.now() - self.active_jobs[0].start_time
+                deadline = self.job_timeout - deadline.total_seconds()
+
+                # Always wait at least 1 second
+                deadline = max(1.0, deadline)
```
enzbang: Should we really put a 1s delay here?

PR #53, opened 2017-02-14
Improve download_file error handling

Nikokrock:
* move exception handler outside contextlib to catch connection error
* add handling of requests.Request as url
* handle missing 'content-disposition' headers
* add tests to cover module

### DIFF
--- a/e3/net/http.py
+++ b/e3/net/http.py
@@ -5,12 +5,15 @@ import contextlib
 import os
 import socket
 import tempfile
+from collections import deque
 
 import e3.log
 import requests
 import requests.adapters
 import requests.exceptions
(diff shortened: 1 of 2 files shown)


e3/net/http.py
```diff
@@ -55,9 +78,54 @@ def set_max_retries(self, base_url,
...
+        """Send a request.
+
+        See requests Session.request function.
+
+        The main difference is that several servers are tried in case
+        base_urls has been set.
+        """
+        error_msgs = []
+        for base_url in list(self.base_urls):
+            logger.debug('try with %s' % base_url)
```
enzbang: use , instead of %

PR #61, opened 2017-02-28
kill method of e3.os.process should not wait for process to die

Nikokrock:


Nikokrock:
The issue was caused by use of PIPE for processes output which is not a good idea in that cases

PR #84, opened 2017-04-11
Fix psutil relatd failed test on solaris systems.

piratos:
Psutil module is not installed in solaris systems, the test should be expected to fail.


tests/tests_e3/job/scheduling_test.py
```diff
@@ -90,6 +92,9 @@ def collect(job):
             assert v[0] is None
             assert v[1] is None
 
+    @pytest.mark.xfail(sys.platform.startswith('sunos'),
```
enzbang: I would prefer a test:

try:
    import psutil
except ImportError:
   psutil = None

@pytest.mark.xfail(psutil is None, ...):

Also fix your commit message ;)

PR #112, opened 2017-05-29
Requests > 2.6.1 breaks urllib3's Retry functionality

piratos:
Using lib requests <= 2.6.1 because later version  breaks urllib3's Retry functionality


enzbang:
This has been fixed upstream.

PR #119, opened 2017-06-06
Protect Job instance attributes with a shared lock

### LINKED ISSUE #118: e3.scheduler race condition
There is thread unsafe code in e3.job which makes our TestScheduler::test_timeout fail randomly, see e.g. https://travis-ci.org/AdaCore/e3-core/jobs/239531977

Access to Job attributes that are modified in a thread and read in another one need to be protected with locks. We can start with a single lock shared by all Job instances.

enzbang:
Also make sure that we're not spawning several processes at the
same time since it seems to impact badly the performances.

Fixes #118

### DIFF
--- a/e3/job/__init__.py
+++ b/e3/job/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+from collections import namedtuple
 from datetime import datetime
 from e3.os.process import Run
 import abc
@@ -10,8 +11,12 @@ import threading
 logger = e3.log.getLogger('job')
 
 
+JobTimingInfo = namedtuple('JobTimingInfo',
+                           ['start_time', 'stop_time', 'duration'])
+
+
 class Job(object):
-    """Class to handle a single Job.
+    """Handle a single Job.
 
     :ivar slot: number associated with the job during its execution. At a
         given time only one job have a given slot number.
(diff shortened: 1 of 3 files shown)


e3/job/__init__.py
```diff
@@ -107,8 +144,12 @@ def run(self):
         # Do non blocking spawn followed by a wait in order to have
         # self.proc_handle set. This allows support for interrupt.
         cmd_options['bg'] = True
-        self.proc_handle = Run(self.cmdline, **cmd_options)
-        self.proc_handle.wait()
+        with self._lock:
```
Nikokrock: You cannot avoid the check on self.interrupted here as the following scenario can occurs:
A- (lock) check for interrupted
B- (lock) set to interrupted
C- (lock) run

e3/job/__init__.py
```diff
@@ -55,40 +61,71 @@ def __init__(self, uid, data, notify_end):
...
-        pass
+        pass  # all: no cover
 
     def interrupt(self):
-        """Interrupt current job."""
-        self.interrupted = True
+        """Interrupt current job.
+
+        :rtype: bool
+        :return: True if interrupted, False is already interrupted
```
Nikokrock: is -> if

e3/job/__init__.py
enzbang: Thank you for the review, defensive code added!

PR #125, opened 2017-06-12
Fix failing test due to gitignore file and swallowed exceptions

piratos:
- Electrolyt job should return failure statement if code crashes for any reason
- ( Test fix ) So far we don't test the existence of .gitignore before opening them in the specs


e3/electrolyt/run.py
```diff
@@ -47,8 +47,17 @@ def run(self):
                 method_name = 'do_%s' % class_name.lower()
                 if self.dry_run:
                     return
-                if hasattr(self, method_name):
-                    getattr(self, method_name)()
+                try:
+                    getattr(self, method_name)
```
enzbang: please assign the result to a variable, don't run getattr twice

tests/tests_e3/anod/test_sandbox.py
```diff
@@ -100,6 +101,10 @@ def e3fake_git_dir(git, tmpdir):
 
         if os.path.isdir(os.path.join(e3_fake_git_dir, '.git')):
             return
+        # Add gitignore
+        if not os.path.exists(os.path.join(e3_fake_git_dir, '.gitignore')):
+            touch(os.path.join(e3_fake_git_dir, '.gitignore'))
+
```
enzbang: I don't understand why you need a .gitignore here

PR #129, opened 2017-06-14
Increase statement coverage

enzbang:


e3/anod/context.py
```diff
@@ -403,12 +403,31 @@ def add_spec(self,
         return result
 
     @classmethod
+    def enforce_decision(cls, action, decision):
+        """Force explicit decision."""
```
Nikokrock: Can you add a docstring?

PR #130, opened 2017-06-15
Fix windows tests and a job processing typo

piratos:


enzbang:
What am I supposed to do with this pull-request? As discussed live we don't want to have a test using virtualenv in the testsuite. This is taking a long time for no good reason.

piratos:
i am refoactoring the tests to remove any dependency to virtualenv or e3 specs

PR #132, opened 2017-06-19
Rewrite tests to remove virtualenv and e3 dependecies

piratos:
Now tests are done either on e3fake which is a dependency-free clone
of e3 or an empty dummyspec.


piratos:
need to cover anod action in electrolyt_run

PR #139, opened 2017-07-28
Fixes for upload in PyPI

enzbang:
Note that we might want to switch README.md to rst in order to have a good display on PyPI.

See https://pypi.org/project/e3-core/


enzbang:
Another solution would be to write a shorter text redirecting to GitHub. This would probably be better than trying to have a README file suitable for the two different usages.

PR #140, opened 2017-09-02
Minor cleanup

enzbang:


e3/os/process.py
```diff
@@ -487,14 +487,17 @@ def poll(self):
         else:
             return None
 
-    def kill(self, recursive=True):
+    def kill(self, recursive=True, timeout=3):
         """Kill the process.
 
         :param recursive: if True, try to kill the complete process tree
         :type recursive: bool
+        :param timeout: wait timeout after sending the kill signal (when
```
brobecke: Document the unit used for the timeout? (seconds?)

tests/tests_e3/os/process/main_test.py
```diff
@@ -228,45 +229,62 @@ def test_interrupt():
...
+                with open(pid_file) as f:
+                    child_pid = f.read()
+                    if child_pid:
+                        break
+            except IOError:
+                pass
             time.sleep(0.1)
-        assert len(p_one_child_children) == 1
-        return p_one_child, p_one_child_children
+        e3.fs.rm(gen_prog_name)
```
brobecke: Isn't this file still locked because the process you spawned is still running?

e3/os/process.py
enzbang: I can indeed document the unit.

tests/tests_e3/os/process/main_test.py
enzbang: No it's not. Or else the check would not have passed on AppVeyor ;)

PR #148, opened 2017-10-19
Review use of priorities in e3.job.scheduler

Nikokrock:
1- by default all jobs have the same priority
2- ensure stability of the queue when two jobs have the same priority

### DIFF
--- a/e3/job/__init__.py
+++ b/e3/job/__init__.py
@@ -77,18 +77,15 @@ class Job(object):
             self.index = Job.index_counter
             Job.index_counter += 1
 
-    def __lt__(self, other):
-        """Compare two jobs.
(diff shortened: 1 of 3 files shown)


e3/job/__init__.py
```diff
@@ -81,14 +81,23 @@ def __lt__(self, other):
...
         :rtype: bool
         """
-        return self.index < other.index
+        return False
+
+    def __eq__(self, other):
+        """Check for equality.
+
+        Implementation is needed in order to be able to compare tuples
+        containing Job objects.
```
enzbang: That's too strange ;)

As discussed live, introduce a priority attribute and use it in e3.job.scheduler.

PR #173, opened 2018-02-01
Review patch discarding mechanism

Nikokrock:
Fix issue where Ada line start with '-- ' was matched as the start
of a patch.


tox.ini
```diff
@@ -31,7 +31,7 @@ deps =
       pycodestyle>=2.0.0,!=2.1.0
       pyflakes>=1.2.3
       pydocstyle>=1.0.0
-      isort
+      isort!=4.3.0
```
enzbang: please also change the commit log ;)

PR #174, opened 2018-02-01
Minor production adjustments

Nikokrock:
* Remove unnecessary requirements in tox
* Ensure isort version 3.4.0 is not used
* Add testing on x86_64-windows

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,7 @@
 version: 1.0.{build}
 environment:
   PYTHON: C:\\Python27
+  PYTHON64: C:\\Python27-x64
 install:
 - cmd: >-
     SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -11,7 +12,25 @@ install:
...
+- cmd: >-
+    SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
+
+    tox -e py-ci-cov-codecov
+
+    if %errorlevel% neq 0 exit /b %errorlevel%
+
+    SET PATH=%PYTHON64%;%PYTHON64%\\Scripts;%PATH%
+
+    tox -e py-ci-cov-codecov
```
enzbang: Run without codecov in 64bit mode. This is not working and I don't think we need it.

Nikokrock: so tox -e py ?

PR #175, opened 2018-02-02
New class e3.job.walk.Walk to walk through a DAG

brobecke:
Hello,

Here is an updated version of this effort.

Fingerprints: The core of the fingerprint is an OrderedDict. As far as I can tell from testing and reading/googling around, json.dump should be preserving the order in the output, and then json.load can be configured to load it back preserving that order too. So I think it should be OK. However, it's possible we might want to change the OrderedDict for a simple dict instead. I can do that either as a preliminary patch, or as a followup one.

Do we want the "dry run" mode to be integrated in the Walk class itself, or do we want to leave that to child classes? As you can see from the testcase, it is possible to implement that mode in the derived class without too much fuss -- at least for this testcase. However, child classes might appreciate it if we provided an extra parameter for walking the DAG in dry-run mode. The downside is that it complexifies this code. The upside is that every child classes gets it for free. I vote for adding it in class Walk.

There is a fair amount of testing; not sure if it is worth testing more scenarios or not: At this point, I think adding more tests is mostly validating the deriving of the Walk classes (hence our testcase), rather than the real non-testing code. I have a feeling that the exhaustive scenario testing will have to be done when we test the real user of this class (electrolyt, IIUC), so I don't think we really need to do more here. The current testing has a benefit, however, of showing we implement the features I know we will eventually need. Did I miss any feature?

There is also something that we may need to clarify, and pay special attention to. It's the notion of "forced_skip" vs "skip". I interpreted "forced_skip" as being the dry-run mode - i.e. an action that should have been run, but will be replaced by something that does nothing and return success instead. And "skip" is the result of an action that we decided via should_execute_action that we did not need to run. I tried to implement this class based on that understanding.


enzbang:
Why do we need ordered dict? Looking quickly at the fingerprint implementation I don't see a real need. Let's try to simplify that part.

Also what would be the use case for a dry-run feature here? In our context this will not work as the fingerprint can depend on binaries that will be downloaded. So a dry-run would not work in that context. Unless we have a strong use case in mind I would suggest not implementing dry run now. We can always add it later on.

brobecke:
Re: Fingerprint data stored inside an OrderedDict: Agreed. I will change it to use a simple dict, and adjust whatever needs to be adjusted to provide stable output/fingerprint computation.

Re: dry-run. Good points. For now, I will keep the concept in the tests I wrote, as it shows we can do something along those lines, but will keep it outside of the main framework.

PR #176, opened 2018-02-14
Move the Fingerprint class from e3.anod hierarchy to e3

brobecke:


brobecke:
Ping? Can someone review this patch, please?

PR #178, opened 2018-02-20
Add read and verify (JW)Token

AnnaKaczorowski:
TN:R115-030


tests/tests_e3/net/token/main_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+TESTS_E3_DIR = os.path.dirname(os.path.dirname(os.path.dirname(
+    os.path.abspath(__file__))))
+TOKEN_PATH = os.path.join(TESTS_E3_DIR, 'data', 'token_file')
+FUTURE_TIMESTAMP = 9999999999999
+
+
+def test_valid_token():
+    payload = {u'typ': u'Bearer', u'exp': FUTURE_TIMESTAMP}
+    if is_valid(payload):
```
Nikokrock: assert is_valid(payload)

e3/net/token.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def utc_timestamp():
+    """Return the number of seconds since epoch UTC."""
+    d = datetime.utcnow()
+    return calendar.timegm(d.utctimetuple())
+
+
+def is_valid(token):
+    """Return true if the token is still valid, false otherwise."""
```
Nikokrock: use proper docstring

e3/net/token.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # Add required padding
+    rem = len(payload) % 4
+    if rem > 0:
+        payload += b'=' * (4 - rem)
+    return jwt, json.loads(base64.b64decode(payload).decode('utf-8'))
+
+
+def utc_timestamp():
+    """Return the number of seconds since epoch UTC."""
```
Nikokrock: use proper docstring. It helps tools such pycharm

e3/net/token.py
Nikokrock: technically that's token_payload for the parameter not the full token. what about accepting here a full token and doing the decoding in the is_valid function ? The token might not always come from a file

PR #179, opened 2018-03-04
enhance class Anod.BuildSpace to reuse class Fingerprint's load/save methods

brobecke:
All in the subject and in the commits' revision log.

Small note: This is unrelated to this pull request, but I noticed that BuildSpace.load_fingerprint has a "sha_only" option, where the current behavior is to load the contents of the file and return its sha1. Not sure if this is supposed to be correlated to class Fingerprint's sha1 method. It's something different, right?

### DIFF
--- a/e3/anod/buildspace.py
+++ b/e3/anod/buildspace.py
@@ -9,14 +9,12 @@ from datetime import datetime as dt
 import e3.error
 import e3.log
 import e3.os.process
-import yaml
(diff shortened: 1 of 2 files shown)


enzbang:
Hi Joel,

Looking at our previous implementation of build space (before e3) it seems that sha1_only is never set to True. I suggest that we remove this option now.

PR #184, opened 2018-03-29
Change python_script for .exe scripts

AnnaKaczorowski:
If python shebang is in the exe script return interpreter path too.

### DIFF
--- a/e3/sys.py
+++ b/e3/sys.py
@@ -235,6 +235,20 @@ def python_script(name, prefix=None):
     :return: a list that will be the prefix of your command line
     :rtype: list[str]
     """
+    def has_relative_python_shebang(file_script):  # unix: no cover
+        """Return True if the script contains #!python shebang.
+
+        When producing relocatable python distribution we change the shebang
+        to #!python. In that case prefix the command line with the current
+        python interpreter.
+
+        #!/path/to/python shebang should return false, as we don't need to
+        return interpreter path.
+        """
+        with open(file_script, 'rb') as f:
+            content = f.read()
+        return re.search(b'#!python', content, flags=re.MULTILINE) is not None
+
     if prefix is None:
         if sys.platform == 'win32':  # unix: no cover
             prefix = os.path.dirname(sys.executable)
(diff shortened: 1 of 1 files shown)


e3/sys.py
```diff
@@ -235,18 +235,40 @@ def python_script(name, prefix=None):
     :return: a list that will be the prefix of your command line
     :rtype: list[str]
     """
+    def python_shebang(file_script):  # unix: no cover
+        """Return True if the script contains #!python shebang.
+
+        note: '#!/path/to/python' should return false, as we don't need to
```
enzbang: When producing relocatable python distribution we change the shebang to #!python. In that case prefix the command line with the current python interpreter.

e3/sys.py
```diff
@@ -235,18 +235,44 @@ def python_script(name, prefix=None):
...
+
     if prefix is None:
         if sys.platform == 'win32':  # unix: no cover
             prefix = os.path.dirname(sys.executable)
         else:
             prefix = os.path.dirname(os.path.dirname(sys.executable))
 
     if sys.platform == 'win32':  # unix: no cover
-        script = os.path.join(prefix, 'Scripts', name)
+        if os.path.basename(prefix) == 'scripts':  # defensive code
```
enzbang: I don't like seeing #defensive code on two branches. Can you rewrite that as:

script = os.path.join(prefix, name) if os.path.basename(prefix) == 'scripts' \
    else os.path.join(prefix, 'Scripts', name)

e3/sys.py
```diff
@@ -235,18 +235,44 @@ def python_script(name, prefix=None):
...
+            # probably a virtualenv
+            script = os.path.join(prefix, name)
+        else:  # defensive code
+            script = os.path.join(prefix, 'Scripts', name)
+
+        script_exe = script + '.exe'
         if script.endswith('exe') and os.path.isfile(script):
+            if has_relative_python_shebang(script):  # defensive code
```
enzbang: use` # all: no cover` instead of ` # defensive code` here

e3/sys.py
```diff
@@ -235,18 +235,44 @@ def python_script(name, prefix=None):
...
+                # relocatable python distribution
+                return [interpreter(prefix), script]
             return [script]
-        elif os.path.isfile(script + '.exe'):
-            return [script + '.exe']
+        elif os.path.isfile(script_exe):
+            if has_relative_python_shebang(script_exe):  # defensive code
```
enzbang: use` # all: no cover` instead of ` # defensive code` here

PR #210, opened 2018-09-17
Add SVN high lvl interface (for source packaging)

AnnaKaczorowski:


e3/vcs/svn.py
```diff
@@ -0,0 +1,166 @@
...
+        :type url: str
+        :param revision: specific revision (default is last)
+        :type revision: str | None
+        :param force_and_clean: if True: erase the content of non empty
+        working_copy and use '--force' option for the svn update/checkout
+        command
+        :type force_and_clean: bool
+        :raise: SVNError
+        """
+        def is_clean_svn_dir(dir_path, ignore_change=True):
```
enzbang: Having the ignore_change in that function makes the logic a little too complex IMHO. I would prefer to remove it here and always return True, False when a svn dir is not clean. Then you can perfectly ignore the "not clean" status when not in force_and_clean.

PR #216, opened 2018-10-04
Add possibility to disable lines in electrolyt plans

enzbang:
Using:

   with defaults(enabled=False):
        ...

disable all lines in the "with" block


tests/tests_e3/electrolyt/plan/main_test.py
```diff
@@ -161,3 +161,24 @@ def test_verify_entry_point():
...
+        content=[u'def myserver():\n',
+                 u'    build("foo")\n'
+                 u'    with defaults(enabled=False):\n'
+                 u'        build("bar1")\n'
+                 # It is not possible to "revert" the enabled=False status
+                 u'        with defaults(enabled=True):\n'
+                 u'            build("bar2")\n'
+                 u'    build("bar3")\n'])
+
+    actions = context.execute(myplan, 'myserver', verify=False)
```
brobecke: So, IIUC, calling execute with verify=True, we should be getting 'foo', 'bar1', 'bar2', and 'bar3', right? Should we add a test for that?

e3/electrolyt/plan.py
```diff
@@ -92,14 +99,23 @@ def __init__(self,
...
         :type kwargs: dict
         """
+        self.ignore_disabled = ignore_disabled
         if stack:
             # This is a scope creation, so copy current env
             self.stack = stack
             self.plan = plan
             new = self.stack[-1].copy(build=build, host=host, target=target)
+
+            if new.enabled and not enabled:
```
brobecke: IIUC, the expected way to use this is to initially create a context with stack=None, and then it will recursively take care of  creating the stack elements; as a result, the BaseEnv objets in the stack should always have an "enabled" attribute. But someone calling this initialization method with a stack he created himself for whatever reason would get an exception, here, because of this attribute being missing. Maybe the easiest is just to document the fact that the BaseEnv object has an additional attribute?

enzbang: We already document that stack is "internal". I'm improving the doc to explicitly say that user instantiation of PlanContext should be done with stack set to None.

tests/tests_e3/electrolyt/plan/main_test.py
enzbang: No, you have this result when you instantiate PlanContext with ignore_disabled=False, I've added a test.

PR #224, opened 2018-11-20
SVN update returns status about local changes

AnnaKaczorowski:


e3/vcs/svn.py
```diff
@@ -164,3 +166,4 @@ def is_empty_dir(dir_path):
 
         mkdir(self.working_copy)
         self.svn_cmd(['checkout', url, '.'] + options)
+        return not (not is_svn_dir or is_clean)
```
enzbang: Hum... if is_svn_dir is False you should return True. There is a local chage.

PR #254, opened 2019-03-25
Add prune function on DAGs

Nikokrock:


e3/collection/dag.py
```diff
@@ -492,6 +492,41 @@ def as_dot(self):
         result.append("}")
         return "\n".join(result)
 
+    def prune(self, fun):
+        """Create a pruned graph.
+
+        Note that tags are preserved but get_context calls might change
```
enzbang: That's too dangerous IMHO to keep some tags and discard others. get_context might return wrong data in some cases but still work in most of the case and debugging that part might be difficult.

I would prefer an additional parameter preserve_tags that will never prune a tag if set to True and that will not keep any tag if set to False

PR #258, opened 2019-04-16
Add function to check if a given path is a supported archive format

Nikokrock:


e3/archive.py
```diff
@@ -22,6 +22,18 @@ class ArchiveError(e3.error.E3Error):
     pass
 
 
+def is_archive_supported(filename):
```
enzbang: is_known_archive_format? is_supported_archive_format? can_unpack?

PR #265, opened 2019-05-28
Various improvements and bug fixes

Nikokrock:


e3/vcs/svn.py
```diff
@@ -107,7 +107,11 @@ def current_revision(self):
         :rtype: str
         :raise: SVNError
         """
-        return self.get_info('Last Changed Rev')
+        try:
+            return self.get_info('Last Changed Rev')
+        except Exception as e:
+            logger.exception("Cannot fetch last changed rev")
+            raise SVNError("Cannot fetch last changed rev", "svn_cmd")
```
brobecke: Can we include the exception message as well? It might contain some information helping diagnose the issue...

e3/fingerprint.py
```diff
@@ -225,12 +225,16 @@ def load_from_file(cls, filename):
...
             - The file does not contain a fingerprint we recognize;
             - The fingerprint has a version number we do not support.
+            - If the fingerprint file does not exist
 
         :param filename: The name of the file where to load the fingerprint
             from.
         :type filename: str
         :rtype: Fingerprint
         """
+        if not os.path.isfile(filename):
```
brobecke: This change means we can simplify the "load_previous_fingerprint" method in class ElectrolytWalk:

``        if os.path.exists(filename):
            return Fingerprint.load_from_file(filename)
        else:
            return None
``

e3/anod/context.py
```diff
@@ -342,6 +344,17 @@ def add_spec(self,
         :type source_name: str | None
         :param force_source_deps: whether to force loading the source deps
         :type force_source_deps: bool
+        :param plan_line: corresponding line:linenumber in the plan
+        :type plan_line: str
+        :param plan_args: action args after plan execution, taking into
+            account plan context (such as with defaults(XXX):)
+        :type plan_args: dict
+        :param sandbox: if not None, anod instance are automatically bind to the
```
brobecke:
```suggestion
        :param sandbox: if not None, anod instance are automatically bound to the
```

e3/anod/context.py
```diff
@@ -342,6 +344,17 @@ def add_spec(self,
...
+            given sandbox
+        :type sandbox: None | Sandbox
+        :param upload: if True consider uploads to the store (sources and
```
brobecke: What does "consider" mean, in practice?

e3/anod/status.py
```diff
@@ -14,3 +14,4 @@ class ReturnValue(Enum):
     force_fail = 123
     unknown = 124
     skip = 125
+    unchanged = 126
```
brobecke: This new enum seems unused?

e3/electrolyt/run.py
```diff
@@ -171,6 +171,11 @@ def do_uploadbinarycomponent(self):
         # not implemented
         self.__status = STATUS.success
 
+    def do_uploadsource(self):
```
brobecke: This doesn't seem to be used anywhere. Does it need to be added here?

PR #279, opened 2019-07-18
Add source closure resolution

Nikokrock:


e3/anod/queries.py
```diff
@@ -0,0 +1,156 @@
...
+        """Compute the source closure (internal function).
+
+        :param spec: an anod instance to inspect
+        :type spec: Anod
+        :param publish: whether the sources found should be marked internal
+            or not.
+        :type publish: bool
+        """
+        # If this is an installation add the corresponding package
+        if spec.kind == 'install':
```
enzbang: are we sure to always have an uploaded package in that case?

e3/anod/spec.py
```diff
@@ -144,7 +144,10 @@ def __init__(self, qualifier, kind, jobs=1, env=None):
         :raise: SpecError
         """
         self.deps = OrderedDict()
-        """:type: OrderedDict[e3.anod.deps.Dependency]"""
+        """:type: OrderedDict[str, e3.anod.Anod]"""
+
+        self.dependencies = OrderedDict()
```
enzbang: This is a significant change that change the API. Either we dump the API number and document the change or we try to move this info somewhere else.

PR #284, opened 2019-07-23
Fix issues with Windows version detection

Nikokrock:
* Newer versions of Python might now detect correctly Windows 2016
* Use build numer to make distinction between 2016 and 2019

Part of RA31-005


enzbang:
Already merged in stable

PR #286, opened 2019-08-02
Fail when a repository does not exist

enzbang:
Instead of generating a failure when the configuration for a
repository is not found it is better to crash completely when loading
the DAG. This will help detecting issues early during plan checks.

TN: S701-019

### DIFF
--- a/e3/electrolyt/run.py
+++ b/e3/electrolyt/run.py
@@ -100,10 +100,6 @@ class ElectrolytJob(Job):
     def do_checkout(self):
         """Get sources from vcs to sandbox vcs_dir."""
         repo_name = self.data.repo_name
-        if self.data.repo_data is None:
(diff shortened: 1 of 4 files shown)


e3/electrolyt/run.py
```diff
@@ -100,7 +100,7 @@ def do_test(self):
     def do_checkout(self):
         """Get sources from vcs to sandbox vcs_dir."""
         repo_name = self.data.repo_name
-        if self.data.repo_data is None:
+        if self.data.repo_data is None:  # defensive code
```
brobecke: Just wondering, why not just assert, here, since the change in e3.anod.context should guaranty that this is no longer possible, right?

enzbang: I've remove this code. We do not even need an assert.

PR #287, opened 2019-08-02
Rejects explict calls to install() when build() is needed

enzbang:
TN: S628-019

### DIFF
--- a/e3/anod/context.py
+++ b/e3/anod/context.py
@@ -422,6 +422,12 @@ class AnodContext(object):
 
         if primitive == 'install' and not spec.has_package and \
                 has_primitive(spec, 'build'):
+            if plan_line is not None and plan_args is not None:
+                # We have an explicit call to install() in the plan but the
+                # spec has no binary package to download.
+                raise SchedulingError(
+                    "error in plan at {}: "
+                    "install should be replaced by build".format(plan_line))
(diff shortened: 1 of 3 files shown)


e3/anod/context.py
```diff
@@ -422,6 +422,12 @@ def add_action(data, connect_with=None):
 
         if primitive == 'install' and not spec.has_package and \
                 has_primitive(spec, 'build'):
+            if plan_line is not None and plan_args is not None:
+                # We have an explicit call to install() in the plan but the
+                # spec has no binary package to download.
+                raise SchedulingError(
+                    "add_spec error: install primitive should be changed to"
+                    " a build for the plan_line {}".format(plan_line))
```
brobecke: FWIW, putting myself in the shoes of an inexperienced developer, the message reads a bit confusing to me; trying to think about it, I think it's a few tiny details:
  - "add_spec" is foreign to the user. Is it really necessary in this case?
 - "install primitive": it's not the primitive that should be changed, but the call. I know it sounds a bit pedantic, but it got me wondering for a while.

Howzabout: "Error in plan at: {}\n" + "install should be replaced by build"; or
"Error in plan (install should be replaced by build) at: {}"

?

enzbang: Good point. I've submitted a new change

PR #288, opened 2019-09-03
Reject duplicate deps

enzbang:


e3/anod/context.py
```diff
@@ -364,6 +364,25 @@ def add_action(data, connect_with=None):
             if connect_with is not None:
                 self.connect(connect_with, data)
 
+        def add_dep(deps, dep, dep_instance):
+            """Add a new dependency in a dependencies dict.
+
+            :param deps: the Anod instance ``deps`` attribute
```
brobecke: FWIW, one thing that's missing in the documentation for me is which dependency's list is being updated.
We could solve that problem very simply by just adding that piece of info in this description. On the other hand, how about passing the spec whose dependencies we want to update instead? For me, I think itll be easier to document, and easier I think to read the code... But if you prefer keeping the code, as is, perhaps change the description to:

    The ``deps`` attribute of *an* Anod instance, to which the new dependency is to be added.

e3/anod/context.py
```diff
@@ -364,6 +364,25 @@ def add_action(data, connect_with=None):
...
+            :type deps: dict
+            :param dep: a Dependency instance
```
brobecke: suggest, for clarity's sake, that this is the dependency we want to add to the dependency list?

e3/anod/context.py
```diff
@@ -364,6 +364,25 @@ def add_action(data, connect_with=None):
...
+            :param dep_instance: the Anod instance loaded for that dependency
+            :type dep_instance: Anod
+            """
+            if dep.local_name in deps:
+                raise AnodError(
+                    origin='expand_spec',
+                    message='Error when adding dependency '
+                    '({}, local_name: {}): '
+                    'A dependency with the same local_name '
+                    'has already been loaded'.format(dep.name, dep.local_name))
```
brobecke: Thinking from a user's perspective, what we mean to say is that a spec has two Dependency objects with the same local_name attribute, right? If correct, perhaps we could phrase the error message using that perspective, rather than from the point of view of the mechanics of how/when this was detected?

e3/anod/context.py
```diff
@@ -422,6 +422,12 @@ def add_action(data, connect_with=None):
 
         if primitive == 'install' and not spec.has_package and \
                 has_primitive(spec, 'build'):
+            if plan_line is not None and plan_args is not None:
+                # We have an explicit call to install() in the plan but the
+                # spec has no binary package to download.
+                raise SchedulingError(
+                    "add_spec error: install primitive should be changed to"
+                    " a build for the plan_line {}".format(plan_line))
```
brobecke: See suggestion on this error message in pull request #287

PR #300, opened 2019-09-20
Adjust default implementation of Package to make it generic

Nikokrock:


brobecke:
Thanks for doing that Nico. This hadn't occurred to me until this morning.

I think this fixes AdaCore's S919-054. Can you add that to the subject of your commit?

PR #304, opened 2019-09-24
Add class to manage checkout done with Anod

Nikokrock:


e3/vcs/git.py
```diff
@@ -63,6 +64,29 @@ def __init__(self, working_tree):
...
+        :param repo_path: a local directory that contains the repository
+        :type repo_path: str
+        :param initial_content_path: directory containing the initial content
+            of the repository. If set to None an empty repository is created.
+        :type initial_content_path: str
+        :return: the URL of the newly created repository
+        :rtype: str
+        """
+        repo_path = os.path.abspath(repo_path)
+        repo = GitRepository(repo_path)
```
brobecke: Could we use "cls" instead of hardcoding "GitRepository" here? That way, if we happen to change the class' name, we wouldn't have to change it here.

e3/vcs/svn.py
```diff
@@ -46,6 +46,54 @@ def __init__(self, working_copy):
         """
         self.working_copy = working_copy
 
+    @classmethod
+    def local_url(self, repo_path):
```
brobecke: For classmethod, let's not call this parameter "self"?

e3/vcs/svn.py
```diff
@@ -46,6 +46,54 @@ def __init__(self, working_copy):
...
+                repo_path = repo_path[0].upper() + repo_path[1:]
+            return 'file:///' + repo_path.replace('\\', '/')
+        else:  # windows: no cover
+            return 'file://' + repo_path
+
+    @classmethod
+    def create(cls, repo_path, initial_content_path=None):
+        """Create a local subversion repository.
+
+        :param repo_path: a local directory that contains the repository
```
brobecke: The wording suggests, at least to me, that repo_path is a directory that already contains the SVN repo. But reading the code, it's actually the location where the repostory will be created, right?

So perhaps just change the documentation to say: "A local directory where to create the repository" ?

e3/vcs/svn.py
```diff
@@ -46,6 +46,54 @@ def __init__(self, working_copy):
...
+        :type repo_path: str
+        :param initial_content_path: directory containing the initial content
+            of the repository. If set to None an empty repository is created.
+        :type initial_content_path: str
```
brobecke: -> "str | None"

e3/vcs/git.py
```diff
@@ -63,6 +64,29 @@ def __init__(self, working_tree):
...
 
+    @classmethod
+    def create(cls, repo_path, initial_content_path=None):
+        """Create a local Git repository.
+
+        :param repo_path: a local directory that contains the repository
+        :type repo_path: str
+        :param initial_content_path: directory containing the initial content
+            of the repository. If set to None an empty repository is created.
+        :type initial_content_path: str
```
brobecke: -> str | None

e3/vcs/svn.py
```diff
@@ -89,6 +137,8 @@ def get_info(self, item):
         """
         info = self.svn_cmd(['info'], output=PIPE).out
         m = re.search(r'^{item}: *(.*)\n'.format(item=item), info, flags=re.M)
+        if m is None:
```
brobecke: Other than the logging, I don't see any specific handling of that case, meaning we'll get to line 142, which should result in a crash because 'm' is None. Can we add a test in the testsuite to check for that?

e3/vcs/svn.py
```diff
@@ -46,6 +46,54 @@ def __init__(self, working_copy):
...
+        if sys.platform == 'win32':  # unix: no cover
+            if len(repo_path) > 1 and repo_path[1] == ':':
+                # svn info returns the URL with an uppercase letter drive
+                repo_path = repo_path[0].upper() + repo_path[1:]
+            return 'file:///' + repo_path.replace('\\', '/')
+        else:  # windows: no cover
+            return 'file://' + repo_path
+
+    @classmethod
+    def create(cls, repo_path, initial_content_path=None):
```
brobecke: Food for thought. I just realized while reading the testcases that, with SVN, the "create" method creates a different kind of SVN repository -- in other words, the word "repository" is overloaded; it could me the central copy that users check out from and push to (what this method creates), vs the "working copy" (what this class works with).

I think it's worth enhancing the documentation to be clearer about this.

PR #305, opened 2019-09-26
Improve documentation of e3.vcs

Nikokrock:


e3/vcs/svn.py
```diff
@@ -68,11 +68,16 @@ def local_url(self, repo_path):
     def create(cls, repo_path, initial_content_path=None):
         """Create a local subversion repository.
 
-        :param repo_path: a local directory that contains the repository
+        This create a local repository (not a working copy) that can be
```
brobecke: create -> creates

e3/vcs/svn.py
```diff
@@ -68,11 +68,16 @@ def local_url(self, repo_path):
...
+        referenced by using file:// protocol. The purpose of the this function
+        is mainly to test svn related function without relying on a remote
```
brobecke: "svn related" -> "svn-related"
"function" -> "functions"

e3/vcs/svn.py
Nikokrock: Fixed

e3/vcs/svn.py
Nikokrock: Fixed

PR #306, opened 2019-09-26
Handle checkout fingerprints

Nikokrock:


e3/anod/checkout.py
```diff
@@ -31,6 +31,34 @@ def __init__(self, name, working_dir):
...
+        if is_prediction:
+            return None
+
+        try:
+            with open(self.metadata_file) as fd:
+                vcs_info = json.load(fd)
+            fingerprint.add(
+                self.name, (vcs_info['url'],
+                            vcs_info['new_commit'],
+                            vcs_info['revision']))
```
enzbang: revision should not be needed, double check for subversion

enzbang: I think we should use `git ls-tree -r HEAD` for Git instead of looking at url and commit id. This is really what we care about for fingerprint.

e3/anod/checkout.py
```diff
@@ -31,6 +31,33 @@ def __init__(self, name, working_dir):
...
+
+        try:
+            with open(self.metadata_file) as fd:
+                vcs_info = json.load(fd)
+            fingerprint.add(
+                self.name, (vcs_info['url'],
+                            vcs_info['new_commit']))
+            return fingerprint
+        except Exception:
+            logger.exception('Got exception')
```
brobecke: I think we will find it very useful for analysis purposes to log the exception itself.

e3/anod/checkout.py
Nikokrock: Added for git only

e3/anod/checkout.py
Nikokrock: the call to logger.exception will output the traceback and all the relevant exception information. Regarding the interface, I am returning the fingerprint as in some cases we return None (which is in the end a special fingerprint value). The use case for this code will be to call it from the walker to compute the fingerprint.

brobecke: Re: logger.exception: Ok, good.

Still don't understand all that well the use case, but I suppose this will become clearer when I see it actually become used.

No further comment on this pull request.

enzbang:
Can you give us a status update? Is that still needed?

PR #308, opened 2019-10-01
Fix issue with wrong interpreter used.

Nikokrock:
When parse_shebang is enabled on Windows, the python interpreter
used to run e3.os.process is taken by default instead of the python
in the path. Ensure that which command is invoked to avoid the issue


brobecke:
Shouldn't we have a test to verify that we're getting the behavior we want?

PR #309, opened 2019-10-18
Add support for external repositories

Nikokrock:


e3/anod/checkout.py
```diff
@@ -60,6 +63,36 @@ def update(self, vcs, url, revision=None):
...
+        :param revision: ignored
+        :type revision: None
+        """
+        # Expand env variables and ~
+        url = os.path.expandvars(os.path.expanduser(url))
+
+        old_commit = get_filetree_state(self.working_dir)
+        ignore_list = []
+        if os.path.isfile(os.path.join(url, '.gitignore')):
+            with open(os.path.join(url, '.gitignore')) as fd:
```
enzbang: as discussed live you can use

git ls-files -o --ignored --exclude-standard

PR #310, opened 2019-10-23
Ensure file permissions are preserved when unzipping files.

Nikokrock:
The file permissions are only preserved on non-Windows systems where
they are relevant (executable bit for example)

### DIFF
--- a/e3/archive.py
+++ b/e3/archive.py
@@ -18,6 +18,28 @@ import e3.os.fs
 logger = e3.log.getLogger('archive')
 
 
+class E3ZipFile(zipfile.ZipFile):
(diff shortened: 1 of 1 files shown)


e3/archive.py
```diff
@@ -18,6 +18,25 @@
...
+
+        if sys.platform != 'win32':
+            # Try to preserve attributes on non Windows platforms as
+            # executable attribute is relevant on those platforms. As we rely
+            # on an internal ignore any errors at this stage.
+            try:
+                attr = member.external_attr >> 16
+                if attr != 0:
+                    os.chmod(result, attr)
+            except Exception:
```
enzbang: AttributeErrror

PR #323, opened 2020-01-08
improve error message when a plan has an install that should be a build

brobecke:
The new error message gives a bit more information, and tries to provide
more possible solutions to the issue, hoping that this will help users
with less experience around Anod specs and electrolyt plans.

Seen while working on T107-003.

Additional note that's not in this commit's revision log: Please feel free to suggest a better error message!

### DIFF
--- a/e3/anod/context.py
+++ b/e3/anod/context.py
@@ -438,7 +438,10 @@ class AnodContext(object):
                 # spec has no binary package to download.
                 raise SchedulingError(
                     "error in plan at {}: "
-                    "install should be replaced by build".format(plan_line))
+                    "Cannot install `{}' because this spec does not produce "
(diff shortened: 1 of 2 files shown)


e3/anod/context.py
```diff
@@ -438,7 +438,10 @@ def add_dep(spec_instance, dep, dep_instance):
                 # spec has no binary package to download.
                 raise SchedulingError(
                     "error in plan at {}: "
-                    "install should be replaced by build".format(plan_line))
+                    "Cannot install `{}' because this spec does not produce "
+                    "a package; check you are using the appropriate "
```
enzbang: this is misleading: it is not that the spec does not define a package, it is that the spec does the define a way of downloading a binary package.

PR #330, opened 2020-01-17
Avoid collision on s3 bucket receiving events

Nikokrock:
A same events can be sent several times with a different status
(running, failed, ...).

### DIFF
--- a/e3/event/__init__.py
+++ b/e3/event/__init__.py
@@ -18,6 +18,17 @@ from e3.fs import mkdir
 logger = e3.log.getLogger('event')
 
 
+def unique_id():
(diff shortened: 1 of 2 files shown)


e3/event/__init__.py
```diff
@@ -18,6 +18,15 @@
 logger = e3.log.getLogger('event')
 
 
+def unique_id():
+    """Return a random globally unique id.
+
+    :return: an id
+    :rtype: str
+    """
+    return str(uuid.uuid1(clock_seq=int(1000 * time.time())))
```
enzbang: Add a comment to explain the use of clock_seq

PR #339, opened 2020-01-31
Improve plan checker error when missing explicit build

enzbang:
When the plan has a build_tree dependency on a spec and the resolver
configuration require an explicit build in the plan mention the two
possible alternatives:

- add the explicit build in the plan
- or modify the dependency on the spec

TN: SA22-006

### DIFF
--- a/e3/anod/context.py
+++ b/e3/anod/context.py
@@ -642,8 +642,18 @@ class AnodContext(object):
         """
         if decision.choice is None and decision.expected_choice in (
                 Decision.LEFT, Decision.RIGHT):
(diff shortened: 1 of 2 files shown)


enzbang:
> Thanks for making error message more detailed and hopefully useful for users!
> 
> The change looks good to me. Just one possible thought (not sure about it): Would it be worth suggesting the change in the dependency first, with the idea being that an install dependency is a 
> bit more flexible and therefore more desirable, and so something we want to suggest first, so people try that first?

I'm no so sure about that given that changing the dependency can introduce a bug, adding the explicit build line is safe so I think it makes more sense to add it first.

PR #344, opened 2020-02-14
Add new BuildSpace.exists method

brobecke:
This method allows users to determine whether a build space
(still?) exists or not. One possible application of that method
is in electrolyt, where we can take the existance of an action's
build space (or lack thereof) into consideration when deciding
whether to execute that action or not.

TN: S114-025


brobecke:
I will send an updated patch with additional testing...

PR #347, opened 2020-02-20
Handle diff not encoded as utf-8 and add ability to disable diff computation when managing a repository

Nikokrock:


e3/text.py
```diff
@@ -20,3 +20,26 @@ def format_with_dict(pattern, values):
     key_regexp = r"|".join([r'\(%s\)' % k for k in values])
     result = re.sub(r'%%(?!%s)' % key_regexp, r'%%', pattern)
     return result % values
+
+
+def bytes_as_str(content):
+    """Safely convert bytes in a python string.
```
enzbang: to

e3/text.py
```diff
@@ -20,3 +20,26 @@ def format_with_dict(pattern, values):
...
+
+    The conversion function attempt to convert to decode utf-8 and
+    in case of decoding error fallback on a safe representation in
```
enzbang: fall back

brobecke:
The changes themselves look reasonable to me, but I don't understand why we are not
adding some tests for these. Is it because we are pressed for time? Or because we feel
this isn't necessary in this case. I would agree with going fast if it is pressing (but with
a note and a TN to finish this later), but I would argue against the "not worth it" reason.

e3/text.py
```diff
@@ -20,3 +20,26 @@ def format_with_dict(pattern, values):
     key_regexp = r"|".join([r'\(%s\)' % k for k in values])
     result = re.sub(r'%%(?!%s)' % key_regexp, r'%%', pattern)
     return result % values
+
+
+def bytes_as_str(content):
+    """Safely convert bytes to a python string.
+
+    The conversion function attempt to convert to decode utf-8 and
```
brobecke: A bit hard to process, at least for the first line... How about?

This function attempts the conversion assuming a utf-8 enconding.
If this triggers a conversion error, it then falls back to a safe representation,
where some characters might be escaped.

PR #348, opened 2020-02-26
Ensure unzip preserves executable attributes

Nikokrock:


e3/archive.py
```diff
@@ -32,6 +32,8 @@ def _extract_member(self, member, path, pwd):
                 # preserve bits 0-8 only: rwxrwxrwx
                 # this come from a proposed patch on python.org
                 # see: https://bugs.python.org/issue15795
```
brobecke: I think this comment is meant to be attached to the code on line 37. Move the new hunk before this comment?

PR #349, opened 2020-02-26
Adjust process spawn test that use ignore_environ=True

Nikokrock:
On Windows not all environment variables can be discarded.

### DIFF
--- a/e3/os/process.py
+++ b/e3/os/process.py
@@ -302,12 +302,26 @@ class Run(object):
         self.err = ''
         self.cmds = []
 
-        if env is not None and not ignore_environ:
-            # ignore_environ is False, so get a copy of the current
-            # environment and update it with the env dictionnary.
-            tmp = os.environ.copy()
-            tmp.update(env)
-            env = tmp
+        if env is not None:
+            if ignore_environ:
+                if sys.platform == 'win32':
+                    # On Windows not all environment variables can be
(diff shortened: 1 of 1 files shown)


e3/os/process.py
```diff
@@ -302,12 +302,21 @@ def add_interpreter_command(cmd_line):
...
-            # ignore_environ is False, so get a copy of the current
-            # environment and update it with the env dictionnary.
-            tmp = os.environ.copy()
-            tmp.update(env)
-            env = tmp
+        if env is not None:
+            if ignore_environ:
+                if sys.platform == 'win32':
+                    tmp = {}
+                    for var in ('SYSTEMDRIVE', 'SYSTEMROOT'):
```
enzbang: Can you add a comment to explain why?

e3/os/process.py
```diff
@@ -302,12 +302,21 @@ def add_interpreter_command(cmd_line):
...
-        if env is not None and not ignore_environ:
-            # ignore_environ is False, so get a copy of the current
-            # environment and update it with the env dictionnary.
-            tmp = os.environ.copy()
-            tmp.update(env)
-            env = tmp
+        if env is not None:
+            if ignore_environ:
+                if sys.platform == 'win32':
+                    tmp = {}
```
brobecke: I would move the (excellent!) comment you provided in the revision directly in the code, somewhere here.

PR #358, opened 2020-03-27
Sanitize Env.set_env input to handle bad user input

enzbang:
When a user pass the same value for build and host (or host and target)
Env.set_env currently creates different Platform objects and then
checks such as:

    if env.build != env.host: ...

have a surprising behaviour for users.

Let's handle this case in set_env, which is called when reading a
plan.

TN: R502-041

### DIFF
--- a/src/e3/env.py
+++ b/src/e3/env.py
@@ -240,6 +240,14 @@ class AbstractBaseEnv(object, metaclass=abc.ABCMeta):
         :param host: string as passed to --host
         :param target: string as passed to --target
         """
+        # Sanitize the input
+        if host is not None and target == host:
+            # Set target to None if it should inherit the value from host
+            target = None
+        if build is not None and host == build:
+            # Set host to None if it should inherit the value from build
+            host = None
+
         saved_build = self.build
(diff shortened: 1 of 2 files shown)


Nikokrock:
The current behaviour allows all kind of combination regarding the build/host/target. In some cases it might be handy. For example you want to build a canadian x86_64_linux compiler that target rhES5 x86_64-linux.

If you build from an ubuntu16:
--build=x86_64-linux,ubuntu16 --host=x86_64-linux,rhES5

Doing the same thing on a redhat 5.x will result in 
--build=x86_64-linux,rhES5  --host=x86_64-linux,rhES5

If we change the current behaviour the former examples won't behave the same in the sense that one will be a "canadian" whereas the other not. 

Also if only --host is passed the (ie: build=None) the behaviour on a redhat 5.x host is still using the current way. Which is even more confusing I think

The current implementation might not be always intuitive but it is sound (FMHO).

Note that e3 provides helpers to compute the right command line based on a given environment (cmd_triplet, str_triplet, ...). and that the user can use also --host=build for example.

PR #361, opened 2020-04-03
Conditional constants

enzbang:


src/e3/electrolyt/plan.py
```diff
@@ -21,6 +23,57 @@ class PlanError(E3Error):
...
+            anod_build("my_component")
+
+    In that case tools parsing the plan won't be able to detect potential issues.
+    Instead use:
+
+        with defaults(enabled="my_const"):
+             anod_build("my_component")
+    """
+
+    RECORD: List[ConditionalConstant] = []
```
Nikokrock, line 43: I think it would be better to introduce ConstantSet class and move RECORD and shuffle in it

tests/tests_e3/electrolyt/plan/main_test.py
```diff
@@ -175,3 +175,113 @@ def test_plan_disable_lines():
...
+    assert actions2[1].spec == "bar"
+
+    assert not actions2[1].plan_args["enabled"]
+
+    list(plan.ConditionalConstant.shuffle())
+    assert actions2[1].plan_args["enabled"]
+
+
+def test_conditional_constants():
+    plan.ConditionalConstant.RECORD = []
```
Nikokrock, line 212: Code reorganisation will avoid that

src/e3/electrolyt/plan.py
```diff
@@ -64,6 +117,20 @@ def __init__(
         for name, cls in entry_point_cls.items():
             self.mod.__dict__[name] = partial(entry_point, self.entry_points, cls, name)
 
+        self.plan_date = datetime.now(timezone.utc)
+        self.mod.__dict__["cond"] = self.cond
+
+    def cond(self, name: str, date: Callable[[datetime], bool]) -> ConditionalConstant:
```
Nikokrock, line 123: Can you adjust a bit the code to accept for date a boolean ? We might still need that. In that case you may rename the parameter

src/e3/electrolyt/plan.py
```diff
@@ -21,6 +23,57 @@ class PlanError(E3Error):
...
+
+        In some cases we want to evaluate the plan for different set of values
+        for the conditional constants to verify that the plan will still be
+        valid under different conditions.
+        """
+        initial_values = tuple(bool(c) for c in cls.RECORD)
+        for shuffeld_values in itertools.product(
+            [True, False], repeat=len(initial_values)
+        ):
+            if shuffeld_values != initial_values:
```
brobecke, line 70: FWIW, it looks like there is a typo -> "shuffled"

enzbang: indeed! I'll fix it

src/e3/electrolyt/plan.py
enzbang: As discussed live we don't know whether there is a need for now. So we'll keep the code as it is.

src/e3/electrolyt/plan.py
enzbang: Ok, I'll try.

enzbang:
Give the changes I'm doing I'm closing this pull-request, replaced by #365

PR #363, opened 2020-04-05
Use py3 features

enzbang:
Modernize e3-core code:

- remove some kwargs hacks that are not needed anymore
- recommend functools.lru_cache instead of e3.decorator.memoize
- address deprecation warnings

### DIFF
--- a/src/e3/vcs/git.py
+++ b/src/e3/vcs/git.py
@@ -22,26 +22,31 @@ import os
 import sys
 import tempfile
 from contextlib import closing
-from subprocess import PIPE
 from typing import TYPE_CHECKING
 
 import e3.fs
 import e3.log
 import e3.os.fs
 import e3.os.process
+from e3.os.process import PIPE
 from e3.text import bytes_as_str
 from e3.vcs import VCSError
 
-logger = e3.log.getLogger("vcs.git")
-
-HEAD = "HEAD"
-FETCH_HEAD = "FETCH_HEAD"
-
 if TYPE_CHECKING:
-    from typing import Dict, Iterator, IO, List, Optional, TextIO, Union
-    from e3.os.process import Run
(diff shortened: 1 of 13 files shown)


src/e3/vcs/git.py
```diff
@@ -92,25 +95,34 @@ def create(cls, repo_path: str, initial_content_path: Optional[str] = None) -> s
             repo.git_cmd(["commit", "-m", "initial content"])
         return repo_path
 
-    def git_cmd(self, cmd: Git_Cmd, **kwargs) -> Run:
+    def git_cmd(
+        self,
+        cmd: Git_Cmd,
+        output: Union[
+            STDOUT_VALUE, DEVNULL_VALUE, PIPE_VALUE, GIT_LOG_STREAM_VALUE, str, IO, None
```
Nikokrock, line 102: Can't we create an intermediate type Run_Output and in this command have Union[Run_Output, GIT_LOG_STREAM_VALUE]

enzbang: I'm not really convinced that this is a good idea, what we have here is useful I think for documenting the accepted values

src/e3/vcs/git.py
```diff
@@ -22,26 +22,29 @@
...
 if TYPE_CHECKING:
-    from typing import Dict, Iterator, IO, List, Optional, TextIO, Union
-    from e3.os.process import Run
+    from typing import Dict, Final, Iterator, IO, List, Literal, Optional, TextIO, Union
+    from e3.os.process import Run, STDOUT_VALUE, DEVNULL_VALUE, PIPE_VALUE
 
     Git_Cmd = List[Optional[str]]
+    GIT_LOG_STREAM_VALUE = Literal[-4]
+
+GIT_LOG_STREAM: GIT_LOG_STREAM_VALUE = -4
```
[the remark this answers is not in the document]
enzbang: well, it's lower than -3

PR #364, opened 2020-04-06
Add tracking of previous fingerprints.

Nikokrock:
This will be used to implement a "skip update" mechanism using
the compute_fingerprint method.

### DIFF
--- a/src/e3/job/walk.py
+++ b/src/e3/job/walk.py
@@ -21,20 +21,22 @@ class Walk:
 
     :ivar actions: DAG of actions to perform.
     :vartype actions: DAG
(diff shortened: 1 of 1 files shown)


src/e3/job/walk.py
```diff
@@ -21,20 +21,22 @@ class Walk:
 
     :ivar actions: DAG of actions to perform.
     :vartype actions: DAG
+    :ivar prev_fingerprints: A dict of e3.fingerprint.Fingerprint objects,
+          indexed by the corresponding job ID. This dictionary contains
+          the former fingerprints a given job or None if there was no
```
brobecke, line 26: I had to look at the code again to answer the following question, so I suggest we explain what "former" means: Shen is it populated, and as a result, at which point can deriving classes start using that information?

PR #367, opened 2020-04-09
Prepare unreachable code support

Nikokrock:
Adjust code in order to be able to enable warn_unreachable in mypi. Remaining errors at this stage are

src/e3/electrolyt/plan.py:149: error: Right operand of 'and' is never evaluated
src/e3/electrolyt/plan.py:152: error: Statement is unreachable

### DIFF
--- a/src/e3/os/platform.py
+++ b/src/e3/os/platform.py
@@ -45,8 +45,15 @@ class SystemInfo:
 
     # Cache for SystemInfo methods
     _platform: Optional[str] = None
+
+    # _os_version is a tuple: os version, kernel version
     _os_version: Optional[Tuple[str, str]] = None
+
+    # When _is_virtual is True it means the host is a VM
     _is_virtual: Optional[bool] = None
+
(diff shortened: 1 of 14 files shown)


src/e3/os/platform.py
```diff
@@ -44,10 +44,10 @@ class SystemInfo:
...
 
     # Cache for SystemInfo methods
-    _platform = None
-    _os_version = None
-    _is_virtual = None
-    _hostname = None
+    _platform: Optional[str] = None
+    _os_version: Optional[Tuple[str, str]] = None
+    _is_virtual: Optional[bool] = None
+    _hostname: Optional[Tuple[str, str]] = None
```
brobecke, line 50: FWIW, it would have been nice to take this opportunity to document those class variables as well, particularly the ones that are tuples, because it's less obvious (at least to me) why we need tuples.

enzbang:
I've submitted #370 and #369 that should fix the failure you see, rebase your pull-request on top of these commits

PR #375, opened 2020-04-20
Introduce e3 configuration file

enzbang:
e3 config file is based on the toml format, to use it you need to
first choose a specific section, e.g. [log] for e3.log and define
a dataclass and will contain the configuration

@dataclass
class LogConfig:
    # Define default format for StreamHandler and FileHandler
    pretty: bool = True
    stream_fmt: str = "%(levelname)-8s %(message)s"
    file_fmt: str = "%(asctime)s: %(name)-24s: %(levelname)-8s %(message)s"

Then calling config_log = Config.get("log", LogConfig) will return the log
configuration.

If the log file is:

[log]
pretty = False

config_log.pretty will be set to True while config_log.stream_fmt and
config_log.file_fmt will keep their default values.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -43,3 +43,9 @@ ignore_missing_imports = True
 
 [mypy-tqdm.*]
 ignore_missing_imports = True
+
+[mypy-tomlkit.*]
+ignore_missing_imports = True
+
+[mypy-typeguard.*]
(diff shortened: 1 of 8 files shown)


enzbang:
Not ready yet, cannot call get_type_hints in e3.config as it will not have the visibility on all type hints

PR #376, opened 2020-04-20
Introduce e3 configuration file

enzbang:
e3 config file is based on the toml format, to use it you need to
first choose a specific section, e.g. [log] for e3.log and define
a dataclass and will contain the configuration

@dataclass
class LogConfig(ConfigSection):
    title = "log"

    # Define default format for StreamHandler and FileHandler
    pretty: bool = True
    stream_fmt: str = "%(levelname)-8s %(message)s"
    file_fmt: str = "%(asctime)s: %(name)-24s: %(levelname)-8s %(message)s"

Then calling config_log = Config.get("log", LogConfig) will return the log
configuration.

If the log file is:

[log]
pretty = False

config_log.pretty will be set to True while config_log.stream_fmt and
config_log.file_fmt will keep their default values.

### DIFF
--- a/src/e3/config.py
+++ b/src/e3/config.py
@@ -1,0 +1,142 @@
+"""Read e3 config file."""
+from __future__ import annotations
+from dataclasses import fields, dataclass
+
+from typing import TYPE_CHECKING, get_type_hints, ClassVar
+
+try:
+    from typeguard import check_type
+
+    CONFIG_CHECK_TYPE = True
+except ImportError:  # defensive code
+    CONFIG_CHECK_TYPE = False
+
+
+import logging
+import os
+
+if TYPE_CHECKING:
+    from typing import Type, TypeVar
+
+    T = TypeVar("T", bound="ConfigSection")
+
+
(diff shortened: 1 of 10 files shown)


src/e3/config.py
```diff
@@ -0,0 +1,142 @@
...
+        """Load the configuration file(s).
+
+        Note that this method is automatically loaded the first time .get()
+        is called.
+
+        :param filename: if not None load this configuration file instead of
+            the default config files
+        """
+        for config_file in KNOWN_CONFIG_FILES:
+            if os.path.exists(config_file):
```
Nikokrock, line 141: isfile rather than exists

tests/conftest.py
```diff
@@ -13,6 +13,7 @@ def init_testsuite_env():
     # Force UTC timezone
     os.environ["TZ"] = "UTC"
     os.environ["E3_ENABLE_FEATURE"] = "smtp_ssl"
+    os.environ["E3_CONFIG"] = "/dev/null"
```
Nikokrock, line 16: does it works on windows ?

PR #382, opened 2020-05-14
Add support to emit JSON log

camilo1729:
A new formatter was added which produces JSON logs. To activate it
the parameter --json-log is used.

Custom attributes can be added using the method add_formatter_attr
and they can be used using the keywork 'extra':

    logger.debug("message!!", extra={'size': '100'})


camilo1729:
dropping this proposition

PR #386, opened 2020-05-20
Add support for propagating JSON logs to subprocess

camilo1729:
The JSON logs can be activated using:

    export E3_ENABLE_FEATURE=json-logs

This will enable JSON logs in all subprocesses that use the
--console-logs or --log-file options.

### DIFF
--- a/src/e3/main.py
+++ b/src/e3/main.py
@@ -105,17 +105,17 @@ class Main:
             help="disable color and progress bars",
         )
         log_group.add_argument(
-            "--json-log",
-            default=False,
+            "--json-logs",
+            default="json-logs" in os.environ.get("E3_ENABLE_FEATURE", "").split(","),
             action="store_true",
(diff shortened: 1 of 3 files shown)


src/e3/main.py
```diff
@@ -105,7 +105,7 @@ def __init__(self, name: Optional[str] = None, platform_args: bool = False):
             help="disable color and progress bars",
         )
         log_group.add_argument(
-            "--json-log",
+            "--json-logs",
```
enzbang, line 108: to keep the settings at the same place I would have just changed the default here,

default="json-logs" in os.environ.get("E3_ENABLE_FEATURE", "").split(",")

camilo1729: I see but this will be enable JSON logs for normal stdout, I wanted them to be enabled only when --console-logs is used.

camilo1729: Done

PR #396, opened 2020-06-19
Add small send_message helper to quickly send email

enzbang:
TN: T518-022

### DIFF
--- a/src/e3/net/smtp.py
+++ b/src/e3/net/smtp.py
@@ -2,9 +2,11 @@ from __future__ import annotations
 
 import os
 import smtplib
+from email.message import Message
 
 import e3.log
 import e3.os.process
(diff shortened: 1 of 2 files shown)


src/e3/net/smtp.py
```diff
@@ -112,3 +114,31 @@ def system_sendmail() -> bool:
...
+    """Send an e-mail message.
+
+    :param from_email: the address sending this email (e.g. user@example.com)
+    :param to_emails: A list of addresses to send this email to
+    :param subject: the e-mail's subject
+    :param content: the e-mail's content
+    """
+    msg = Message()
+    msg["To"] = ", ".join(to_emails)
+    msg["From"] = from_email
```
camilo1729, line 135: subject probably missing in message

PR #398, opened 2020-06-22
ReturnValue should inherit from IntEnum

manthonyaiello:
Fix for T622-025

### DIFF
--- a/src/e3/anod/status.py
+++ b/src/e3/anod/status.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-from enum import Enum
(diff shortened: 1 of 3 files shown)


manthonyaiello:
I missed an ugly type error related to this change; need to chase it down. Closing for the moment; may abandon.

manthonyaiello:
Reopening now that I've found and fixed a missing case in Job.status that resolved the type error. (The type error actually indicated the missing case; it's quite likely that the case was not possible, but now it's explicit.)

PR #402, opened 2020-06-26
Improve scheduling error message

enzbang:


mypy.ini
```diff
@@ -23,9 +23,6 @@ ignore_missing_imports = True
 [mypy-ld.*]
 ignore_missing_imports = True
 
-[mypy-netifaces.*]
-ignore_missing_imports = True
-
```
brobecke, line 28: Did you mean to bundle this change within this commit?

enzbang: yes, we are removing netifaces

enzbang: hum... no we need to merge https://github.com/AdaCore/e3-core/pull/400 first and then I'll rebase my change on top of it

PR #403, opened 2020-06-26
Reject untyped function definitions

enzbang:
Add disallow_untyped_defs = True to mypy configuration
and disable it when:

- the cost of adding the type hinting is too high (e.g. windows low-level specific code), or
- the modules are not used in production and meant to be replaced in the future

### DIFF
--- a/src/e3/os/process.py
+++ b/src/e3/os/process.py
@@ -405,12 +405,14 @@ class Run:
                 self.internal = Popen(self.cmds[0], **popen_args)
 
             else:
-                runs: List[Popen] = []
+                runs: List[subprocess.Popen] = []
                 for index, cmd in enumerate(self.cmds):
                     if index == 0:
-                        stdin = self.input_file.fd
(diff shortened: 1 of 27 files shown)


src/e3/os/process.py
```diff
@@ -601,7 +601,7 @@ def is_running(self) -> bool:
         else:
             return self.internal.is_running()
 
-    def children(self) -> List[psutil.Process]:
+    def children(self) -> List[Any]:
```
brobecke, line 604: I wish we had a way to get a reminder to check this again once psutil.Process gets type hinting... But I can't think of any way other than having a TODO item to periodically do a review.

Could we perhaps declare a `TypeHintingMissing` type equal to Any, and use that instead? That way, at least, it would be easy to grep for them, and find all the locations where we used `Any` because type hinting was missing. What do people think?

enzbang: I doubt psutil will have type annotation in the near future

PR #407, opened 2020-07-07
Sandbox root dir

enzbang:


src/e3/archive.py
```diff
@@ -20,7 +20,10 @@
 
 
 if TYPE_CHECKING:
-    from typing import Callable, List, Optional, Sequence
+    from typing import Callable, List, Literal, Optional, Sequence, Union
+
+    UnpackAutoRemoveDirType = Literal["auto"]
```
brobecke, line 25: This doesn't seem to be used in either this commit or the next one. But perhaps you meant to use in the declaration of RemoveRootDirType?

src/e3/anod/sandbox/__init__.py
```diff
@@ -40,21 +39,45 @@ def __init__(self) -> None:
...
-        self.meta_dir: Optional[str] = None
-        self.tmp_dir: Optional[str] = None
-        self.tmp_cache_dir: Optional[str] = None
-        self.src_dir: Optional[str] = None
-        self.log_dir: Optional[str] = None
-        self.vcs_dir: Optional[str] = None
-        self.patch_dir: Optional[str] = None
-        self.bin_dir: Optional[str] = None
-        self.__specs_dir: Optional[str] = None
+        self.meta_dir: str = ""
```
brobecke, line 42: Just a question - not a request for change, especially since this would be another change independent from what you are trying to do, here:

Would it make sense, eventually, to pre-initialize those to default directories inside the given root_dir (now that root_dir is going to be mandatory)? One interesting upside of this, I think, is that it would perhaps allow us to get rid of the "dirs" attribute (assuming no one outside the class started using it)?

PR #414, opened 2020-07-27
Adjust url list type

Nikokrock:


src/e3/net/http.py
```diff
@@ -74,7 +74,7 @@ class HTTPSession:
     CHUNK_SIZE = 1024 * 1024
     DEFAULT_TIMEOUT = (60, 60)
 
-    def __init__(self, base_urls: Optional[Union[List[str], List[BaseURL]]] = None):
+    def __init__(self, base_urls: Optional[List[Union[str, BaseURL]]] = None):
```
brobecke, line 77: FWIW, this is hard to read, and would benefit (as a separate change) in declaring a type for URLs, such
as:

URLType = Union[str, BaseURL]

camilo1729: Yes I agree with Joel shouldn’t be hard to include it in this commit 👍

PR #415, opened 2020-08-07
Use git apply in e3.diff.patch method

camilo1729:
Replace patch command by git apply command.

This will allow to handle more functionality like empty files
creation.

Patches using context format are not supported anymore.


enzbang:
Done in  pull request #453

Issue #417, opened 2020-09-01
Instances of `os.rename` should be replaced with `shutil.move`


manthonyaiello:
fix-coverage-paths.py uses `os.rename`, but this is known to fail when the source and destination filesystem are not on the same device. See: https://stackoverflow.com/questions/42392600/oserror-errno-18-invalid-cross-device-link

The suggested fix is to use `shutil.move`, which in python3 correctly handles this case.

manthonyaiello:
This is more pervasive than just the one instance referenced above. Renaming issue accordingly.

PR #418, opened 2020-09-01
Fix issue #417: replace `os.rename`

manthonyaiello:
Instances of `os.rename` in e3 are replaced to use `e3.fs.mv`, which was
itself written to replace `shutil.move`.

Note that in `e3.fs.mv`, the uses of `os.rename` are retained. Quick
testing suggests that these uses do not cause problems, and the error
behavior of `os.rename` is sufficiently different from `shutil.move`
that I am reluctant to change these uses.

In the test case for `e3.fs`, the use of `os.rename` is replaced with
`shutil.move` - it seemed self-serving to use `e3.fs.mv` here.

Finally, in fix-coverage-paths, the use of `os.rename` is replaced with
`shutil.move`. This script doesn't import e3 at all, so sticking with
python libraries seemed best.

### DIFF
--- a/src/e3/anod/sandbox/migrate.py
+++ b/src/e3/anod/sandbox/migrate.py
@@ -40,6 +40,7 @@ def migrate_v1_5() -> None:
     """
     from glob import glob
     import os
+    from e3.fs import mv
 
     for f in glob("*.yaml"):
         print("looking at %s" % f)
@@ -50,7 +51,7 @@ def migrate_v1_5() -> None:
             except OSError:
                 pass
             try:
-                os.rename(name + ".yaml", os.path.join(name, "config.yaml"))
+                mv(name + ".yaml", os.path.join(name, "config.yaml"))
             except OSError:  # defensive code
                 print("error for %s" % name)
         elif "-" in name:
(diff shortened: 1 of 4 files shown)


manthonyaiello:
CI failed running black for files I didn't modify. This is odd:
1. it passed locally when I ran both pre-commit and tox
2. it’s failing on files I didn’t touch
3. the way it’s failing is particularly bizarre — it’s like black is trying to actually edit the file, even though it’s supposed to be running in --diff --check mode

Let me know what you want me to do about this (if anything)

manthonyaiello:
Number (3) above (which reports in the log as "I/O operation on closed file") is a reported issue on black: https://github.com/psf/black/issues/1664

I'm struggling understand why black changed the formatting for these files at all. In the case of archive.py, it reformatted 2/5 lines that were longer than 79 characters — but the two reformatted lines are 87 characters long (prior to reformatting), which it should have left alone. This feels like a black bug.

Perhaps we should be freezing the version of black we use for checking style?

enzbang:
> Perhaps we should be freezing the version of black we use for checking style? E.g., 19.10b0 certainly works (matches the current formatting and doesn't block up with the I/O operation error).

We're trying to follow the latest versions of all dependencies in the GitHub CI, this allow us to quickly detect potential issues way before we deploy in production. In that case, I have freeze to 19.10b0 to avoid these annoying regressions. We'll try again to remove that constraint once those issues are fixed.

enzbang (approved):

PR #423, opened 2020-09-15
Force subclassing SpecConfig to pass data to prolog.py

enzbang:
Maintaining code based on e3-core is much easier if data are well typed. SpecConfig being on type Any makes it harder to verify that there is no obvious bug in the application. Also having a clearer interface helps teams maintaining anod specification files.

Note that this is a slight backward incompatible change.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 * Minor backward incompatible changes:
   * the discrimiant ``is_virtual`` has been removed
   * e3.anod.sandbox.SandBox now has a mandatory root_dir attribute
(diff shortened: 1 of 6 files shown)


Nikokrock (changes requested):
Can you add tests with real repositories.yaml and override ?

Nikokrock (approved):

manthonyaiello:
@enzbang you requested my review, but I don't have triage rights. :) So ... no can do. But consider these changes approved. (github should tell you when you asked for a review from someone who can't give it ....

enzbang:
Thank you @manthonyaiello !

PR #426, opened 2020-09-18
Refactor code: create a subclass PlanActionEnv

enzbang:
Instead of using directly BaseEnv to contain the action environment
(including the name of the action, the qualifier, ...), create a
new subclass PlanActionEnv. This hopefully make the code clearer.

Also change e3.env.BaseEnv type annotations to make it possible to
subclass BaseEnv and keep mypy happy.

add_action_action env parameter is renamed action_env to make it clear
that we're not manipulating BaseEnv objects but PlanActionEnv objects.

TN: T107-002

### DIFF
--- a/src/e3/electrolyt/plan.py
+++ b/src/e3/electrolyt/plan.py
@@ -14,7 +14,7 @@ from e3.error import E3Error
 
 if TYPE_CHECKING:
     from types import TracebackType
-    from typing import Any, Callable, Dict, List, Type, Optional
+    from typing import Any, Callable, Dict, List, Optional, Type
     from e3.collection.toggleable_bool import ToggleableBoolean
     from e3.electrolyt.entry_point import EntryPoint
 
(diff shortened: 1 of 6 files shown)


src/e3/electrolyt/plan.py
```diff
@@ -108,15 +108,28 @@ def load_chunk(self, source_code: bytes, filename: str = "<unknown>") -> None:
...
     """Context in which a Plan is executed."""
 
     def __init__(
         self,
-        stack: Optional[List[BaseEnv]] = None,
+        stack: Optional[List[PlanActionEnv]] = None,
         plan: Optional[Plan] = None,
         ignore_disabled: bool = True,
-        server: Optional[BaseEnv] = None,
+        server: Optional[PlanActionEnv] = None,
```
Nikokrock, line 132: can be BaseEnv here

Nikokrock (changes requested):
Almost ok minus the fact that some BaseEnv are still passed

brobecke (commented):
Thanks for doing that, Olivier. This makes things clearer for me. I'll let @Nikokrock give the final approval.

Nikokrock (approved):

PR #431, opened 2020-09-30
Avoid crash when no predecessors for an install action

enzbang:
In some context, the build tree dependency can produce a subtree
having only an install node. Make sure we don't crash in that case.

TN: T930-046


brobecke (approved):
LGTM. Thanks Olivier.

PR #436, opened 2020-10-07
Disable sockets during tests unless when explicitely needed

enzbang:


brobecke (approved):
Interesting! (if I understand how the fixture works correctly, but I think I do)

PR #437, opened 2020-10-07
Improve DAG implementation and documentation

Nikokrock:
* Document how to speed up insertion by using update_vertex and a
single check
* Improve speed of get_closure (remove quadratic worst case)

### DIFF
--- a/src/e3/collection/dag.py
+++ b/src/e3/collection/dag.py
@@ -311,6 +311,11 @@ class DAG:
     ) -> None:
         """Add a new vertex into the DAG.
 
+        Note that this function always checks that there is no cycle in the
+        current DAG. If you prefer to insert all your node first and then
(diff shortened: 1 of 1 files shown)


src/e3/collection/dag.py
```diff
@@ -506,25 +511,26 @@ def check(self) -> None:
...
+            for n in to_visit:
+                closure.add(n)
+                next_visit |= self.get_predecessors(n)
 
-                closure |= self.get_predecessors(n)
+            to_visit = next_visit - closure
 
-            if len(closure) == closure_len:
+            if len(to_visit) == 0:
                 break
```
JosephMarotte, line 532: It's possible to remove these lines and add it as the condition of the `while` loop

Nikokrock: Sorry miss your message. Indeed the code can be refactored. Missed that

PR #438, opened 2020-10-21
Improve error messages in case of patching error

yakobowski:
The current output is misleading:

    e3.diff.DiffError: patch: running patch -p0 -f < [patchname]
    in [dir] failed with patching file file1
    patching file file2
    [...]
    <error>

It is all too easy to think that the error occurs when patching file1,
especially if the real error is not immediately visible.

### DIFF
--- a/src/e3/diff.py
+++ b/src/e3/diff.py
@@ -111,7 +111,7 @@ def patch(
         if p.status != 0:
             raise DiffError(
                 origin="patch",
-                message="running %s < %s in %s failed with %s"
+                message="running %s < %s in %s failed with:\n%s"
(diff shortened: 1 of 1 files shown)


yakobowski:
@Nikokrock can you review? Thanks.

yakobowski:
Temporarily rebased on top of Olivier's fix to see if I can get a clean Travis build.

enzbang:
@yakobowski you will need my other commit to fix windows tests

yakobowski:
Right. I see a that everything is green now, but I had to leave a bit before you forced-pushed. Let me fix that.

enzbang (approved):

enzbang (approved):

PR #445, opened 2020-11-09
Set 'x86_64-windows64' as default platform for windows

camilo1729:


enzbang (changes requested):
We're not ready to change the default, impact is too big.

enzbang (approved):

PR #446, opened 2020-11-09
Set 'x86_64-windows64' as default platform for windows

camilo1729:
A new entry was added to platform knowledge_base in order to
make 'x86_64-windows64' the default build platform for windows.


enzbang (changes requested):
We're not ready to change the default, impact is too big.

PR #450, opened 2020-11-20
Add ability to add additional s3 prefix that depends on the event.

Nikokrock:
Part of T710-017

### DIFF
--- a/src/e3/event/handler/s3.py
+++ b/src/e3/event/handler/s3.py
@@ -50,6 +50,20 @@ class S3Handler(EventHandler):
             self.aws_profile if self.aws_profile is not None else "",
         )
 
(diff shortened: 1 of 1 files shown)


src/e3/event/handler/s3.py
```diff
@@ -50,6 +50,19 @@ def encode_config(self) -> str:
             self.aws_profile if self.aws_profile is not None else "",
         )
 
+    def s3_prefix(self, event: Event) -> str:
+        """Compute a prefix that will be happened to event_s3_url and log_s3_url.
```
flingure, line 54: I would expect a prefix to be "prepended" instead of "appended" ;)

flingure (approved):

PR #454, opened 2020-12-02
Warn on missing source dependencies

enzbang:
When a source package is used by the spec, the spec needs to have
visibility on its source builder. This can be done either by having
the source builder defined on the same spec, or by adding a "source_pkg"
dependency to the spec defining the source builder.

Also fix the call to set_builder that should take a SourceBuilder
object.

TN: TB05-005


Nikokrock (changes requested):
Even if this is a minor change, can you add a test that covers the warning. TIA

Nikokrock (approved):

PR #457, opened 2020-12-07
Various optimizations to speed up plan checking

pmderodat:


src/e3/anod/context.py
```diff
@@ -611,11 +611,9 @@ def add_dep(spec_instance: Anod, dep: Dependency, dep_instance: Anod) -> None:
 
         # Look for dependencies
         spec_dependencies = []
-        if (
-            f"{primitive}_deps" in dir(spec)
-            and getattr(spec, f"{primitive}_deps") is not None
-        ):
-            spec_dependencies += getattr(spec, f"{primitive}_deps")
+        deps = getattr(spec, f"{primitive}_deps", None)
```
enzbang, line 614: I'm not sure about this one. This hides AttributeError raised in <primitive>_deps. Same in other occurrences below.

That's why usage of dir() was initially done, to raise clear error when the attributes were invalid.

src/e3/collection/dag.py
```diff
@@ -144,6 +144,16 @@ def __init__(self) -> None:
         self.__vertex_predecessors: Dict[VertexID, FrozenSet[VertexID]] = {}
         self.__vertex_successors: Dict[VertexID, FrozenSet[VertexID]] = {}
         self.__has_cycle: Optional[bool] = None
+        self.__cached_topological_order: Optional[List[Tuple[VertexID, Any]]] = None
+
+    def reset_caches(self) -> None:
+        """
```
enzbang, line 150: Make it a valid pep257 docstring: so it should start with `"""Reset caches...`

enzbang (changes requested):
I'll let @Nikokrock review the DAG part. I have already added some comments.

src/e3/anod/context.py
pmderodat: I see, thanks. With all cumulated performance changes, reverting this part makes the time to check plans go from 20s to 26s. I think we can find another way that will not hide these errors.

pmderodat: Done: the new version does not hide AttributeError exceptions.

src/e3/anod/spec.py
```diff
@@ -205,6 +206,20 @@ def __init__(
         # Hold the result of the pre function
         self._pre: Optional[Dict[str, Any]] = None
 
+    def _fetch_attr(self, name: str, default_value: Any) -> Any:
```
enzbang, line 210: Do you think you can move this function outside this class?

I'd like to avoid polluting specs namespace. Btw thank you for taking the time to find this much better solution!

pmderodat: Sure!

Ah yes, this should be done, now, thanks.

src/e3/anod/spec.py
```diff
@@ -78,6 +78,28 @@ def has_primitive(anod_instance: Anod, name: str) -> bool:
...
 
+def fetch_attr(instance: Any, name: str, default_value: Any) -> Any:
+    """Return an attribute or the default value if missing.
+
+    Unlike `getattr(instance, name, default_value)`, this works only on
+    attributes present in the class (so class attributes or properties) and
+    this does not hide AttributeError exceptions that getting an existing
+    attribute might raise.
+    """
+
```
enzbang, line 90: no blank line here, this has been detected by the CI:

`src/e3/anod/spec.py:83:1: D202 No blank lines allowed after function docstring
1027`

You can run it locally with `tox -e checkstyle`

enzbang (commented):
Looks good to me now, just a minor comment detected by the CI

src/e3/collection/dag.py
```diff
@@ -546,19 +561,34 @@ def reverse_graph(self) -> DAG:
...
-            result.check()
-        except DAGError:
-            # Check detected
-            self.__has_cycle = True
-            raise
-        else:
-            # No cycle in the DAG
-            self.__has_cycle = False
+
+        if enable_checks:
```
Nikokrock, line 565: reuse the cache information before doing the checks. Otherwise it will be wrong for two reason: if the original graphs check was not done then you might end with __has_cycle set to None even if you do the check. If the original has has_cycle set to False you will run an unnecessary check.

Nikokrock (changes requested):

src/e3/anod/spec.py
pmderodat: Done, thanks. I proably missed this one because I can’t get a clean `tox` run even with the current `master` branch. FTR, I get this:

```
>                       raise DistributionNotFound(req, requirers)
E                       pkg_resources.DistributionNotFound: The 'netifaces' distribution was not found and is required by the application

/home/pmderodat/_git/e3-core/.tox/py38-cov-xdist/lib/python3.8/site-packages/pkg_resources/__init__.py:770: DistributionNotFound
```
Please let me know if it’s worth opening an issue.

src/e3/collection/dag.py
pmderodat: Done, thanks!

Nikokrock (approved):

src/e3/anod/spec.py
enzbang: As discussed live this was a setup issue on your side and you are now able to run tox.

enzbang (approved):

PR #460, opened 2021-01-06
Test if spec has the requested primitive right after loading it

kokkonisd:
The specs that do not define a source (source_pkg_build) primitive fail
with an unhelpful message, because the `has_primitive()` check was
being done too late.

TN: U106-029


enzbang (commented):
Thank you for your contribution, do you think you can also add a test? We aim for 100% coverage (with exceptions) in e3-core.

Also, one very minor comment: do not add a final dot in the first line of your changelog

enzbang (approved):

PR #462, opened 2021-02-01
Rewrite most union types as X | Y

enzbang:
Take advantage of PEP 604 and mypy 0.800
TN: U127-021

### DIFF
--- a/src/e3/anod/spec.py
+++ b/src/e3/anod/spec.py
@@ -54,9 +54,7 @@ if TYPE_CHECKING:
     SOURCE_PRIMITIVE = Literal["source"]
 
     # Anod Dependency can target a build, install, or source
-    DEPENDENCY_PRIMITIVE = Union[
-        BUILD_PRIMITIVE, INSTALL_PRIMITIVE, SOURCE_PRIMITIVE,
-    ]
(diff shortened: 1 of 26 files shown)


src/e3/anod/spec.py
```diff
@@ -54,9 +54,7 @@
     SOURCE_PRIMITIVE = Literal["source"]
 
     # Anod Dependency can target a build, install, or source
-    DEPENDENCY_PRIMITIVE = Union[
-        BUILD_PRIMITIVE, INSTALL_PRIMITIVE, SOURCE_PRIMITIVE,
-    ]
+    DEPENDENCY_PRIMITIVE = Union[BUILD_PRIMITIVE, INSTALL_PRIMITIVE, SOURCE_PRIMITIVE]
```
camilo1729, line 57: why Union is still here?

camilo1729 (commented):
one small comment

src/e3/anod/spec.py
enzbang: Because it's needed in such cases. That's not a type annotation here.

camilo1729 (approved):

PR #467, opened 2021-02-17
Add a tmp_dir_root argument to unpack_archive

pierretr:
This argument can be used to avoid an overly long temporary path when
extracting an archive to remove its root.

TN: U120-025

### DIFF
--- a/src/e3/archive.py
+++ b/src/e3/archive.py
@@ -133,6 +133,7 @@ def unpack_archive(
     delete: bool = False,
     ignore: Optional[list[str]] = None,
     preserve_timestamps: bool = True,
+    tmp_dir_root: Optional[str] = None,
 ) -> None:
     """Unpack an archive file (.tgz, .tar.gz, .tar or .zip).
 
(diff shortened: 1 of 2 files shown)


src/e3/archive.py
```diff
@@ -159,6 +160,9 @@ def unpack_archive(
     :param preserve_timestamps: if False and remove_root_dir is True, and the
         target directory exists, ensure that updated files get their timestamp
         updated to current time.
+    :param tmp_dir_root: if not None the temporary directory used to
+        extract the archive will be created in tmp_dir_root directory.
+        This only has an effect when remove_root_dir is not False.
```
Nikokrock, line 165: Document behaviour if None

Nikokrock, line 165: when remove_root_dir is True ;-) easier to read

src/e3/archive.py
```diff
@@ -194,9 +198,9 @@ def unpack_archive(
     # If remove_root_dir is set then extract to a temp directory first.
     # Otherwise extract directly to the final destination
     if remove_root_dir:
-        tmp_dest = tempfile.mkdtemp(
-            prefix="", dir=os.path.dirname(os.path.abspath(dest))
-        )
+        if not tmp_dir_root:
```
Nikokrock, line 201: if tmp_dir_root is None (otherwise you catch also "")

Nikokrock (changes requested):

Nikokrock (approved):

PR #474, opened 2021-03-18
Fix loglevel argument choices

pierretr:
A dict was passed as choices but doing so make the resulting loglevel
namespace contain keys and discard values.

TN: U310-015

### DIFF
--- a/src/e3/log.py
+++ b/src/e3/log.py
@@ -31,6 +31,14 @@ if TYPE_CHECKING:
 
     T = TypeVar("T")
 
+LEVELS = {
+    "DEBUG": logging.DEBUG,
+    "INFO": logging.INFO,
+    "WARNING": logging.WARNING,
+    "ERROR": logging.ERROR,
+    "CRITICAL": logging.CRITICAL,
+}
+
 
 @dataclass
 class LogConfig(ConfigSection):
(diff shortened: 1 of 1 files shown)


src/e3/log.py
```diff
@@ -334,13 +334,7 @@ def add_logging_argument_group(
...
         default=default_level,
         help="set the console log level",
-        choices={
-            "DEBUG": logging.DEBUG,
-            "INFO": logging.INFO,
-            "WARNING": logging.WARNING,
-            "ERROR": logging.ERROR,
-            "CRITICAL": logging.CRITICAL,
-        },
+        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
```
enzbang, line 337: That will be hard to maintain, I would put a LEVELS constant and use choices=list(LEVELS.keys())

enzbang (approved):

src/e3/log.py
```diff
@@ -379,7 +381,10 @@ def activate_with_args(args: Namespace, default_level: int = logging.WARNING) ->
     if args.verbose > 0:
         level = default_level - 10 * args.verbose
     else:
-        level = args.loglevel
+        if args.loglevel in LEVELS:
```
Nikokrock, line 384: why do you need the if ? if an issue with mypy an assert is sufficient

pierretr: Humm, I do not understand why mypy comes into play here?

PR #475, opened 2021-03-22
Ensure that all files zipped on win32 have executable bit set

Nikokrock:
This allows better interaction with Cygwin zip.

### DIFF
--- a/src/e3/archive.py
+++ b/src/e3/archive.py
@@ -41,6 +41,20 @@ else:
 
 logger = e3.log.getLogger("archive")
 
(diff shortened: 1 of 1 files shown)


src/e3/archive.py
```diff
@@ -41,6 +41,20 @@
 
 logger = e3.log.getLogger("archive")
 
+if sys.platform == "win32":
+    # On Windows for the executable bit on all files. This ensures that when
```
enzbang, line 45: s/for/force

enzbang (approved):

PR #477, opened 2021-04-16
Add a new anod dependency kind: download

enzbang:
When adding a Dependency with require="download" the plan scheduler
forces an installation of the dependency from the downloaded binary
package. If another anod spec has a require="build_tree" dependency
on the same spec then plan scheduler will raise an error.

### DIFF
--- a/src/e3/anod/context.py
+++ b/src/e3/anod/context.py
@@ -193,7 +193,7 @@ class AnodContext:
         root: Action,
         left: Action,
         right: Action,
-    ) -> None:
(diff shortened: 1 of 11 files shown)


kokkonisd (dismissed):

liaiss (approved):
Ok for me

Nikokrock (changes requested):
Can you add a test where you force download with a spec without explicit download primitive.

Nikokrock (approved):

PR #489, opened 2021-08-03
If PATHEXT variable is not defined on Windows use .exe rather ""

Nikokrock:
For example on github actions PATHEXT is never defined and assuming an
empty PATHEXT value disable automatic .exe completion on which various
testsuite are relying.


camilo1729:
Everything seems ok, there is a typo in the commit message, rather -> rather than or use instead of

camilo1729 (approved):

PR #493, opened 2021-09-01
Use shallow git fetch if changelog is not computed

pierretr:
If git_shallow_fetch e3 feature is set and changelog does not need to be
computed git fetch is done with a depth of 1. This feature can be used
by CIs to make them faster and fix existing issues with large
repositories.

Part of U831-001


enzbang (approved):
I think you should also pass --depth=1 for the first checkout, maybe do so when a specific E3_ENABLE_FEATURE is passed (e.g. the git_shallow_fetch that you have proposed).

This can be done in a follow-up patch.

enzbang (approved):

PR #498, opened 2021-10-14
Fix event handler configuration through env variables

camilo1729:
This commit removes unnecessary '=' at the end of string
configuration.

TN:UA12-039


liaiss:
This pull request is stall from more than 1 year without any progress on this side. Closing it

PR #504, opened 2021-11-23
Use distro instead of ld

atupone:
It seems to me that distro is the new name of ld
ld stopped in 2016 while distro has been recently updated. Latest release on july
and ld is not present in some linux distro

### DIFF
--- a/src/e3/os/platform.py
+++ b/src/e3/os/platform.py
@@ -78,12 +78,12 @@ class SystemInfo:
 
         # Fetch linux distribution info on linux OS
         if cls.uname.system == "Linux":  # linux-only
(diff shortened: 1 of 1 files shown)


atupone:
I'm not signing a license covering just a replace ld with distro

atupone:
setup.py need to be changed too, again replacing ld with distro.
Maybe other need too

grouigrokon:
Is it ok to mark this pull request a duplicate of https://github.com/AdaCore/e3-core/pull/654 ?

Nikokrock:
e3-core is now using distro. Closing PR

PR #506, opened 2022-01-07
Move the tree.py utility to e3-core

kokkonisd:
This tree-printing utility should be stored in e3-core so that it can be
used by all parts of e3.

TN: V107-016

### DIFF
--- a/src/e3/collection/tree.py
+++ b/src/e3/collection/tree.py
@@ -1,0 +1,187 @@
+from typing import List, Optional
+
+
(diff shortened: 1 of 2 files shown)


Nikokrock (changes requested):
Looks that the structure is similar to DAG one ? Do we really need a new structure (can't we just add a pretty pring on DAG ?)

kokkonisd:
Indeed, we could do that. I'll open a new PR.

PR #507, opened 2022-01-11
Use botocore to send events

camilo1729:
In order to not depend on AWS CLI, the S3
Handler now uses botocore to send the
necessary files to S3. This will make
possible to use this implementation on
Lambda.


liaiss:
This pull request is stall from more than 1 year without any progress on this side. Closing it

PR #512, opened 2022-02-15
Fix typo in quickstart guide

peniblec:
I grepped for other instances of `output=[0-9]`, but nothing jumped at me. Let me know if there are other things I should look at.


brobecke (approved):
Thanks for the fix, Kevin.

PR #518, opened 2022-04-29
Support aarch64-darwin

simonjwright:
Update the knowledge base by adding aarch64-darwin entries.

The new entries are  edited copies of the x86_64-darwin entries; but note these possibly-doubtful aspects.

    (BUILD_TARGETS): possible clash with aarch64-ios?
    (HOST_GUESS): yes, uname -m really does return arm64.

### DIFF
--- a/src/e3/platform_db/knowledge_base.py
+++ b/src/e3/platform_db/knowledge_base.py
@@ -77,6 +77,7 @@ OS_INFO: PlatformDBEntry = {
 }
 
 PLATFORM_INFO: PlatformDBEntry = {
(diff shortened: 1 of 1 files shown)


enzbang:
Hi, first thank you for your contribution. A teammate just told me that it is exactly the same as the one that we have just validated https://github.com/AdaCore/e3-core/pull/532 

We missed that one because the CI was failing but it seems that it was just hanging and we should have sent a message. We had a look at your warnings and we believe that this will work well. Don't hesitate to come back to us if this is not the case.

PR #523, opened 2022-08-19
Use `exists` to test for `.git` in checkout.py

### LINKED ISSUE #522: anod checkout update_external should check for .git with `exists`
In `e3.anod.checkout`, we have:

```
if os.path.isdir(os.path.join(url, ".git")):
    # It seems that this is a git repository. Get the list of files to
    # ignore
```

This works as expected for a "normal" clone of a repository, but fails if the clone is a submodule. In the latter case, `.git` is a regular file that "points to" the parent's `.git` directory (or, more precisely, the submodule's place in parent's `.git` directory, like this:

```
~/test-OpenUxAS-as-submodule/OpenUxAS $ cat .git
gitdir: ../.git/modules/OpenUxAS
```

We should use `exists` rather than `isdir`, thus supporting this use case fully.

manthonyaiello:
Fixes #522 

Instead of using `isdir` use `exists` to test for `.git` as a means of
determining that the external to update is a git repository.

### DIFF
--- a/src/e3/anod/checkout.py
+++ b/src/e3/anod/checkout.py
@@ -148,7 +148,7 @@ class CheckoutManager:
             if p.status != 0:
                 raise e3.error.E3Error("rsync failed")
         else:
-            if os.path.isdir(os.path.join(url, ".git")):
+            if os.path.exists(os.path.join(url, ".git")):
                 # It seems that this is a git repository. Get the list of files to
                 # ignore
                 try:


manthonyaiello:
Hmm. My change seems to have impacted coverage. I don't understand why not - I'm happy to add / change tests as appropriate, but will need some help to understand where to make the changes.

src/e3/anod/checkout.py
```diff
@@ -148,7 +148,7 @@ def update_external(
             if p.status != 0:
                 raise e3.error.E3Error("rsync failed")
         else:
-            if os.path.isdir(os.path.join(url, ".git")):
+            if os.path.exists(os.path.join(url, ".git")):
```
eliericha, line 151: Why do you feel this is a better check? I think `isdir` checks that the directory `exists` AND is a directory. So `isdir` seems like a better check to me.

eliericha: Ah apologies. I simply didn't read the original issue 🤦 . Never mind.

eliericha: Actually in that case I suggest adding a comment before this if-statement explaining that `.git` can be a file in the case of submodules because that's a lesser known case.

Nikokrock (approved):

src/e3/anod/checkout.py
manthonyaiello: @eliericha addressed in #524 @Nikokrock closed this before I had a change to address your comment. :)

PR #524, opened 2022-08-22
Add a comment to explain the change for ".git"

manthonyaiello:
Address @eliericha's comment on #523 by adding a comment to explain the logic of the change.

### DIFF
--- a/src/e3/anod/checkout.py
+++ b/src/e3/anod/checkout.py
@@ -148,6 +148,10 @@ class CheckoutManager:
             if p.status != 0:
                 raise e3.error.E3Error("rsync failed")
         else:
+            # Test for a git repository by looking for ".git" in the current
+            # directory - either as a file or as a directory. We check both
+            # because, if a normal git clone, ".git" is a directory; if a
+            # git clone with `git submodule init``, ".git" is a file.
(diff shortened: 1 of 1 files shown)


src/e3/anod/checkout.py
```diff
@@ -148,6 +148,10 @@ def update_external(
             if p.status != 0:
                 raise e3.error.E3Error("rsync failed")
         else:
+            # Test for a git repository by looking for ".git" in the current
+            # directory - either as a file or as a directory. We check both
+            # because, if a normal git clone, ".git" is a directory; if a
+            # git clone with `git submodule init``, ".git" is a file.
```
brobecke, line 154: Small typo (two backquotes after the git command).

brobecke (commented):
Hi Tony. Thanks for the added comment (very useful). Just a small typo.

src/e3/anod/checkout.py
manthonyaiello: Over-eager helpfulness by VS Code, I suspect ("here, I'll add a closing quote, too!"). Thanks - I fixed it. :)

brobecke (approved):

PR #528, opened 2022-09-16
Add an exception in e3.diff.patch function

leocardao:
Change the e3.diff.patch function to raise
an exception if there is no file to patch.

TN: S909-004


enzbang (commented):
Why changing the default of discarded_file? I don't think we should do that, your proposed change will force parsing the patch to look at the files that would be modified every time we run patch, even when there is no pre-processing needed.

enzbang (approved):

PR #530, opened 2022-09-29
Update mypy to 0.971

Nikokrock:


mypy.ini
```diff
@@ -13,6 +13,9 @@ disallow_any_unimported = True
 disallow_subclassing_any = True
 no_implicit_optional = True
 
+# Some functions in e3.log do override without preserving full signature
+disable_error_code = override
```
enzbang, line 17: I would prefer accepting my PR instead https://github.com/AdaCore/e3-core/pull/542 in order to keep that check where it makes sense for us.

liaiss:
No more needed since we transitioned to mypy 1.1.1

PR #533, opened 2022-10-17
Work around stevedore crash with importlib_metadata >= 5.0.0

enzbang:


enzbang:
I still see PR on stevedore to fix compatibility with importlib_metadata 5.0.0. At this stage I think we should wait and temporarily disable the upgrade to importlib_metadata 5.0.0

PR #559, opened 2023-02-22
Add support for SPDX ExternalRef

enzbang:


src/e3/spdx.py
```diff
@@ -399,6 +401,71 @@ class PackageCopyrightText(SPDXEntryMaybeStrMultilines):
...
+class ExternalRefCategory(Enum):
+    """Identify the category of an ExternalRef."""
+
+    security = "SECURITY"
+    package_manager = "PACKAGE-MANAGER"
+    persistent_id = "PERSISTENT-ID"
+    other = "OTHER"
+
+
+@dataclass
```
Nikokrock, line 413: why do you need dataclass here ?

src/e3/spdx.py
```diff
@@ -399,6 +401,71 @@ class PackageCopyrightText(SPDXEntryMaybeStrMultilines):
...
+        self.reference_category = reference_category
+        self.reference_type = reference_type
+        self.reference_locator = reference_locator
+
+    def __str__(self) -> str:
+        return " ".join(
+            (self.reference_category.value, self.reference_type, self.reference_locator)
+        )
+
+    def to_json_dict(self) -> dict[str, dict[str, str]]:
```
Nikokrock, line 443: missing docstring

src/e3/spdx.py
enzbang: I don't

Nikokrock (approved):

PR #561, opened 2023-02-24
Add small interface to NVD API to query CVE impacting a CPE

enzbang:


src/e3/cve.py
```diff
@@ -0,0 +1,125 @@
...
+        results_per_page: int | None = None,
+    ) -> Iterator[CVE]:
+        """Return a list of matching CVE entries.
+
+        :param no_rejected: remove CVE records with the REJECT or Rejected
+            status from API response
+        :param results_per_page: number of results to return for each request,
+             note that it is recommended to keep the default setting
+        """
+        url = f"https://services.nvd.nist.gov/rest/json/cves/2.0?cpeName={cpe_name}"
```
Nikokrock, line 76: don't build parameters by hand. Use params parameter

enzbang: I don't think it works well for key without value, which is what NVD is advertising for some parameters

Nikokrock (approved):

PR #574, opened 2023-03-23
Add raw_out check to prevent None input. IT-81

liaiss:
Prevent the following exception:

Traceback (most recent call last):
  File "/home/test/e3_issue_reproducer.py", line 11, in <module>
    test_outputs(Run(["./main_both"], output="text.txt", error=subprocess.PIPE))
  File "/home/test/e3_issue_reproducer.py", line 5, in test_outputs
    if cmd.out:
  File "/home/test/e3-distrib-current/lib/python3.9/site-packages/e3/os/process.py", line 489, in out
    return bytes_as_str(self.raw_out)
  File "/home/test/e3-distrib-current/lib/python3.9/site-packages/e3/text.py", line 32, in bytes_as_str
    result = content.decode("utf-8")
AttributeError: 'NoneType' object has no attribute 'decode'

Manage raw_out in the same way than raw_err to fix the issue

### DIFF
--- a/src/e3/anod/checkout.py
+++ b/src/e3/anod/checkout.py
@@ -167,14 +167,16 @@ class CheckoutManager:
                         ],
                         output=PIPE,
                     ).out
-                    if ignore_list_lines is None:
-                        ignore_list = []
-                    else:
-                        ignore_list = [
+
+                    ignore_list = (
+                        []
+                        if ignore_list_lines is None
+                        else [
                             f"/{f.strip().rstrip('/')}"
                             for f in ignore_list_lines.splitlines()
                         ]
-                        logger.debug("Ignore in external: %s", ignore_list)
+                    )
+                    logger.debug("Ignore in external: %s", ignore_list)
                 except Exception:  # defensive code
                     # don't crash on exception
                     pass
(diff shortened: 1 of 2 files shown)


tests/tests_e3/os/process/main_test.py
```diff
@@ -17,6 +18,21 @@
     psutil = None
 
 
+def test_run_stdout_stderr(caplog):
+    """Verify that the parse shebang option works."""
```
Helflym, line 22: Wrong description.

tests/tests_e3/os/process/main_test.py
```diff
@@ -17,6 +18,21 @@
...
+    prog_filename = os.path.join(os.getcwd(), "prog")
+    with open(prog_filename, "wb") as f:
+        f.write(b"#!/usr/bin/env python\n")
+        f.write(b"import sys\n")
+        f.write(b'print("stdout", file=sys.stdout)\n')
+        f.write(b'print("stderr", file=sys.stderr)\n')
+    e3.os.fs.chmod("a+x", prog_filename)
+    p = e3.os.process.Run([prog_filename], error=subprocess.PIPE)
+    assert p.out == "stdout\n"
```
Helflym, line 31: You can also assert that `p.err == "stderr"`. That's the purpose of `error=subprocess.PIPE`. Otherwise, it's being redirected to stdout (which is being checked here, but not below).

Helflym (changes requested):
LGTM after these two changes.

Helflym (dismissed):
Thanks for the patch.

pierretr (approved):

PR #578, opened 2023-03-28
Bump Black, Flake8 and MyPy to the latest stable. 

liaiss:
it/org/software-supply-chain/issues#2

### DIFF
--- a/src/e3/env.py
+++ b/src/e3/env.py
@@ -269,7 +269,6 @@ class AbstractBaseEnv(metaclass=abc.ABCMeta):
             elif not propagate_build_info:
                 return Platform.get(*split_value)  # type: ignore
             else:
-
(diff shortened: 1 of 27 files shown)


src/e3/env.py
```diff
@@ -487,7 +486,7 @@ def to_dict(self) -> dict:
             can be used to format string. For example ``Env().target.os.name``
             will appear with the key ``target_os_name``, ...
         """
-        result = {k: v for k, v in self._items()}
+        result = {k: v for k, v in self._items()}  # noqa: C416
```
enzbang, line 489: to be replaced by 
```
result = dict(self._items())
````

liaiss: Done

enzbang (approved):

pierretr (approved):

PR #580, opened 2023-03-30
Fix mypy issue (unused type: ignore)

liaiss:
During the mypy bump of e3-distrib, some unused type: ignore was detected by mypy

it/org/software-supply-chain/issues#2


enzbang:
Missing ending ) in commit message (and MR message)

enzbang (approved):

PR #587, opened 2023-04-15
Set endianness information on aarch64-linux platforms

brobecke:
The following program shows that the endianness is not set when the platform is aarch64-linux:

    | from e3.env import Env
    | Env().set_target("aarch64-linux")
    | print(Env().target.cpu.endian)

Currently, the program currently prints:

    unknown

This fix allows the endianness to be set to the correct value, which is "little".

This change was tested by running this repository's testsuite as well as AdaCore internal e3-adacore testsuite.

### DIFF
--- a/src/e3/platform_db/knowledge_base.py
+++ b/src/e3/platform_db/knowledge_base.py
@@ -80,7 +80,12 @@ PLATFORM_INFO: PlatformDBEntry = {
     "aarch64-darwin": {"cpu": "aarch64", "os": "darwin", "is_hie": False},
     "aarch64-elf": {"cpu": "aarch64", "os": "none", "is_hie": True},
     "aarch64-ios": {"cpu": "aarch64", "os": "ios", "is_hie": False},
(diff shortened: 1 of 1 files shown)


Nikokrock (commented):
This is strange that you need that. aarch64 at the beginning of the file is marked as little-endian

Nikokrock:
@brobecke the issue is not there but in our our private platform db

brobecke:
> @brobecke the issue is not there but in our our private platform db

I think we have some big-endian baremetal runtimes. So for aarch64-elf, it would be unknown.

brobecke:
> I think we have some big-endian baremetal runtimes. So for aarch64-elf, it would be unknown.

FTAOD, what I mean is that I don't think we should assume the endianness is always the same, and thus only rely on it being set at the CPU level. Instead, I think we should set it for each OS, and for platforms where it's undecided, leave it to "unknown".

WDYT?

Nikokrock (changes requested):
endianness is already set in e3-core for the CPU and nowadays all arm are little endian in practice

PR #597, opened 2023-06-15
Fix CTRL-C on a program launched with rlimit

leocardao:
setpgid should be used when the program is running in the background so as not to worry about signal propagation.

Now we detect whether we are a background process or not, and call setpgid accordingly.

IT-465


tools/rlimit/rlimit.c
```diff
@@ -178,10 +182,15 @@ main (int argc, char **argv)
...
-      if (setpgid (0, 0)) {
-        perror ("setpgid");
-        exit (4);
+      console_pid = tcgetpgrp(0);
+      prog_pid = getpgrp();
+
+      /* Check if we are a background process and if so, setpgid */
+      if (console_pid == -1 || prog_pid == -1
+        || (console_pid != prog_pid && setpgid (0, 0) == -1))
+      {
```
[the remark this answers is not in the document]
leocardao: Done

tools/rlimit/rlimit.c
```diff
@@ -72,6 +74,8 @@ reapchild (int nsig)
 {
   int delay;
 
+  (void)nsig;
```
grouigrokon, line 77: I guess this is to prevent warning on unused variable.

tools/rlimit/rlimit.c
```diff
@@ -125,16 +129,26 @@ reapchild (int nsig)
...
 main (int argc, char **argv)
 {
+  int begin;
   sigset_t block_cld;
 
   /* we need at least 3 args */
   if (argc < 3)
     usage ();
 
+  begin = 2;
```
grouigrokon, line 139: Weird logic ... Why 2 ? If you use 0, then you'll have:
`
/* argv[0] = .../rlimit
￼     argv[begin + 1] = seconds
￼     argv[begin + 2] = command
￼     argv[begin + 3] = args */
`
But that's just cosmetic ...
Still, I would not use that variable name. To me, something like `parsed_options`, or `argv_start_index` ... Again, cosmetic, but better for code reviewers ;)

tools/rlimit/rlimit.c
```diff
@@ -179,9 +193,9 @@ main (int argc, char **argv)
       sigprocmask(SIG_UNBLOCK, &block_cld, NULL);
 
       /* child exec the command in a new process group */
-      if (setpgid (0, 0)) {
-        perror ("setpgid");
-        exit (4);
+      if (setpgid (0, 0) == -1) {
```
grouigrokon, line 196: You are right :) `!= 0`, or `== -1` is the right test !

tools/rlimit/rlimit.c
```diff
@@ -202,30 +216,32 @@ main (int argc, char **argv)
...
+      /* If run with --foreground process group, rlimit must be able to
+       * read and write from a tty (not only read => STDIN is a tty).
+       *
        * If only stdout is a terminal there are no issues with not
        * being the foreground process as most terminals are configured
        * with TOSTOP off (so SIGTTOU is only emited in case of terminal
        * settings change.
        */
-      tcsetpgrp(0, getpgid(getpid()));
+      if (foreground)
```
grouigrokon, line 227: Shall we check returned value ?

tests/tests_e3/os/process/main_test.py
```diff
@@ -115,6 +116,123 @@ def run_test():
...
+    try:
+        from ptyprocess import PtyProcess
+    except ImportError:
+        raise ImportError("ptyprocess is needed to run this tests") from None
+
+    # Only a linux test
+    assert sys.platform.startswith("linux"), "This test make sens only on linux"
+
+    script_to_run = """
+from __future__ import annotations
```
grouigrokon, line 167: I do not think this is needed since there are no annotation in this python scriptlet.

tests/tests_e3/os/process/main_test.py
```diff
@@ -115,6 +116,123 @@ def run_test():
...
+    # Test with --foreground
+    os.environ["PS1"] = "$ "
+    p = PtyProcess.spawn(
+        [e3.os.process.get_rlimit(), "--foreground", "30", "bash", "--norc", "-i"],
+        env=os.environ,
+        echo=False,
+    )
+    p.write(b"echo 'test rlimit --foreground'\n", flush=True)
+    time.sleep(2)
+    assert p.read() == b"$ test rlimit --foreground\r\n$ "
```
grouigrokon, line 215: Are you sure about the `\r` ?

tests/tests_e3/os/process/main_test.py
leocardao: Not really necessary but it's good practice so I use it every time.

tests/tests_e3/os/process/main_test.py
leocardao: Yes, ptyprocess use `\r\n` as end of line.

tools/rlimit/rlimit.c
leocardao: I think so, it's an oversight on my part.

grouigrokon (commented):
Is it ok if aix, darwin and solaris executables are not rebuilt ?
All the other points discussed have been fixed, so, appart from those missing builds (which I am not sure are used anymore ?) MR should be ok for me.

tox.ini
```diff
@@ -6,6 +6,7 @@ deps =
      xdist: pytest-xdist[psutil]
      # ??? needs to be added as a dep of e3-core
      requests-cache
+     ptyprocess
```
Nikokrock, line 9: on on linux

tools/rlimit/rlimit.c
```diff
@@ -202,30 +216,34 @@ main (int argc, char **argv)
...
+	exit(4);
+      }
 
       /* Restore SIGTTIN and SIGTTOU signal to their original value
        * in order not to impact children processes behaviour. */
       signal (SIGTTIN, SIG_DFL);
-      signal (SIGTTOU, SIG_DFL);
-      execvp ((const char *) argv[2], (char *const *) &argv[2]);
-      fprintf (stderr, "rlimit: could not run \"%s\": ", argv[2]);
+      if (foreground)
```
Nikokrock, line 235: add a comment here that when non in foreground we keep SIGTTOU to ignore so that child process can write

PR #602, opened 2023-06-22
Amend anod primitive signature

grouigrokon:
Three commits here.
The first is for linter warning removal.
Second is for english typos.
Last but not least is the real work.

As the `post` parameter of the primitives were not called, I added calls to the post function.
Hope this does not break anything ...

### DIFF
--- a/src/e3/anod/driver.py
+++ b/src/e3/anod/driver.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from functools import wraps
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 import e3.log
 import e3.store
@@ -38,7 +38,7 @@ def primitive_check() -> Callable[[F], F]:
 
 
 class AnodDriver:
-    def __init__(self, anod_instance: Anod, store: Store):
+    def __init__(self, anod_instance: Anod, store: Optional[Store]):
         """Initialize the Anod driver for a given Anod instance.
 
         :param anod_instance: an Anod instance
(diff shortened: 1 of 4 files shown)


src/e3/anod/driver.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from functools import wraps
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
```
enzbang, line 4: Do not use Optional but just | None

enzbang (changes requested):
I'd like first to understand what are the issue you're trying to solve before putting too many comments on your PR

We're currently trying to get rid of the post callback so this patch is going into the wrong direction ;) 

It was mostly to implement post=install which is now handled automatically when creating the DAG

src/e3/anod/driver.py
grouigrokon: Ok. It's just that Rust have that `Option` type which is widely used, so I thought it was a good idea to use :) Will do.

grouigrokon:
As for why I did that, it's just that, diving in the anod files, I had warnings that the type passed to `post` was wrong.
I asked @Nikokrock who said having `str | None` would be enough. I kept the `Callable` part for backward compatibility.
If it's about to be removed ... well, no problem :) It was a way for me to go deeper in the processes and code :)

grouigrokon:
As per [this comment](https://github.com/AdaCore/e3-core/pull/602#issuecomment-1603747883), this PR is rejected.

PR #603, opened 2023-06-30
Skip the rlimit_ctrl_c and rlimit_forground tests if you're not using tox

leocardao:
Some tests require ptyprocess to run, but ptyprocess is not included by default with e3, so they always fail in a production environment (without internet).

### DIFF
--- a/tests/tests_e3/os/process/main_test.py
+++ b/tests/tests_e3/os/process/main_test.py
@@ -120,7 +120,10 @@ def test_rlimit():
         e.restore()
 
 
-@pytest.mark.skipif(sys.platform == "win32", reason="A linux test")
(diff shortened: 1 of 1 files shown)


enzbang (changes requested):
you should use pytest.mark.skipif as done to skip the test on windows

leocardao:
@enzbang This is already done.

enzbang:
@leocardao you have not done what I suggested: using pytest.mark.skipif instead of having an early return in the function code. Currently your patch makes the function pass even if there is a bug. Returning SKIP is much clearer.

leocardao:
This pull request is obsolete, as ptyprocess is now referenced in setup.py.

PR #604, opened 2023-07-06
Add the support of hash content for e3.fs.get_filetree_state. ITPE-56

liaiss:


src/e3/fs.py
```diff
@@ -205,6 +209,12 @@ def compute_state(file_path: str) -> bytes:
         )
         return state.encode("utf-8")
 
+    def get_content(file_path: str) -> bytes:
+        with open(file_path, "rb") as f:
+            data = f.read()
```
leocardao, line 214: You can simply return f.read() here

src/e3/fs.py
```diff
@@ -225,8 +235,15 @@ def compute_state(file_path: str) -> bytes:
                 full_path = os.path.join(root, path)
                 result.update(compute_state(full_path))
 
+                if hash_content:
```
leocardao, line 238: The `if hash_content:` block should be placed outside if/else, since it is common to both if and else.

src/e3/fs.py
```diff
@@ -182,15 +182,19 @@ def find(
...
 
     :param path: root path of the file tree to be checked
     :param ignore_hidden: if True (default) then files and directories
         tarting with a dot are ignored.
+    :param hash_content: if True, include the content in the hash.
+
     :return: a hash as a string
 
-    The function will not report changes in the hash if a file is modified
+    The function will not report changes in the hash by default if a file is modified
```
grouigrokon, line 197: To me, the _By default_ should be at the beginning of the phrase : _By default, the function will not report changes in the hash if a file is modified ..._

src/e3/fs.py
grouigrokon: Nope ;) Here the result is updated in a loop, for every file :)

leocardao: Yes, it's true, I hadn't seen that!

tests/tests_e3/fs/main_test.py
```diff
@@ -208,10 +208,19 @@ def test_tree_state():
...
+
     # check that get_filetree_state accept unicode
     state7 = e3.fs.get_filetree_state("toto")
     assert isinstance(state7, str)
 
+    # check that get_filetree_state with hash_content is different that the
+    # previous state
+    state8 = e3.fs.get_filetree_state("toto", hash_content=True)
+    assert isinstance(state8, str)
+    assert state8 != state7
```
grouigrokon, line 222: As the function behavior depends on the path type (directory vs file) should a test over a directory be added too ?

src/e3/fs.py
liaiss: Done

src/e3/fs.py
liaiss: Done

tests/tests_e3/fs/main_test.py
liaiss: Done

tests/tests_e3/fs/main_test.py
```diff
@@ -191,6 +191,7 @@ def test_tree_state():
     time.sleep(2)
 
     e3.os.fs.touch("toto")
+    e3.os.fs.touch("toto2")
```
leocardao, line 194: The `touch` is not necessary here. You never test "toto2" before writing something to it. Since you open it by writing, the file will be created anyway if it doesn't exist.

liaiss: No, here, I create toto2 before to check that we are generating another hash with content of 2 files, this hash will be compared with the first one that don't look at the content (state4 vs state9

tests/tests_e3/fs/main_test.py
```diff
@@ -208,10 +209,28 @@ def test_tree_state():
     state6 = e3.fs.get_filetree_state("toto")
     assert isinstance(state6, str)
 
+    with open("toto", "wb") as f:
+        f.write(b"hello world.")
+
     # check that get_filetree_state accept unicode
     state7 = e3.fs.get_filetree_state("toto")
     assert isinstance(state7, str)
```
grouigrokon, line 217: Should an `assert state7 != state6` be added here ?

liaiss: Maybe I can add state7 == state6 here as hash content is false by default

liaiss: but here I'm not sure it is useful here

liaiss: done

grouigrokon (approved):

PR #606, opened 2023-07-21
Add e3.anod.spec.Anod.deps_source_list

leocardao:
e3.anod.spec.Anod.deps_source_list contain the dependencies source list (any e3.anod.deps.Dependency that require 'source_pkg')

e3.anod.spec.Anod.deps will not reference anymore the source_pkg dependencies.

### DIFF
--- a/src/e3/anod/context.py
+++ b/src/e3/anod/context.py
@@ -32,7 +32,7 @@ from e3.env import BaseEnv
 from e3.error import E3Error
 
 if TYPE_CHECKING:
-    from typing import cast, NoReturn, Optional, Tuple
+    from typing import cast, NoReturn, Optional
     from collections.abc import Callable
     from e3.anod.action import Action
     from e3.anod.package import SourceBuilder
@@ -44,7 +44,7 @@ if TYPE_CHECKING:
     from e3.mypy import assert_never
 
     # spec name, build env, target env, host env, qualifier, kind, source name
-    CacheKeyType = Tuple[
+    CacheKeyType = tuple[
         str, Platform, Platform, Platform, Optional[str], Optional[str], Optional[str]
     ]
     ResolverType = Callable[[Action, Decision], bool]
@@ -439,7 +439,16 @@ class AnodContext:
                         spec_instance.name, dep.local_name
                     ),
                 )
-            spec_instance.deps[dep.local_name] = dep_instance
+            if dep.kind != "source":
+                spec_instance.deps[dep.local_name] = dep_instance
+            else:
+                srcbuild_list = dep_instance.source_pkg_build
+                if not srcbuild_list:
+                    spec_instance.deps_source_list[dep.local_name] = set()
+                else:
+                    spec_instance.deps_source_list[dep.local_name] = {
+                        srcbuild.name for srcbuild in srcbuild_list
+                    }
 
         # Initialize a spec instance
         spec = self.load(
@@ -669,7 +678,7 @@ class AnodContext:
                     add_dep(spec_instance=spec, dep=e, dep_instance=child_instance)
                     self.dependencies[spec.uid][e.local_name] = (
                         e,
-                        spec.deps[e.local_name],
+                        child_instance,
                     )
 
                     continue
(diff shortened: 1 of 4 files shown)


src/e3/anod/context.py
```diff
@@ -439,7 +439,22 @@ def add_dep(spec_instance: Anod, dep: Dependency, dep_instance: Anod) -> None:
...
                 )
-            spec_instance.deps[dep.local_name] = dep_instance
+            if dep.kind != "source":
+                spec_instance.deps[dep.local_name] = dep_instance
+            else:
+                srcbuild_list = dep_instance.source_pkg_build
+                spec_instance.source_list[dep.local_name] = set()
+                if not srcbuild_list:
+                    raise AnodError(
+                        f"{dep.local_name} is marked as source_pkg but not sources is "
```
grouigrokon, line 449: no *source* is defined, ot no sources *are* defined ;)

src/e3/anod/driver.py
```diff
@@ -61,7 +61,28 @@ def activate(self, sandbox: SandBox, spec_repository: AnodSpecRepository) -> Non
                     kind=e.kind,
                     env=e.env(self.anod_instance, BaseEnv.from_env()),
                 )
-                self.anod_instance.deps[e.local_name] = dep_instance
+                if e.kind != "source":
```
grouigrokon, line 64: Hmmm, feel like this could be shared between the two files ? Maybe create a `@staticmethod` that you can call with parameters ?

`def get_source_dependency(e)` then call for `self.anod_instance.deps[e.local_name] = AnodContext.get_source_dependency(e)`

src/e3/anod/context.py
leocardao: `are` seems better, I will correct this typo :)

src/e3/anod/driver.py
leocardao: What do you mean by `shared between the two files`?

grouigrokon: Well to me, the changes here and in `AnodContext` are exactly the same. So the code should be done in one place only ... Still, this is my point of view.

leocardao: No, because AnodDriver is used by the test suite, and is scheduled to be removed. Even if the code seems to be the same, AnodDriver doesn't do the same thing as AnodContext.

grouigrokon: Ok, but you have the error message to fix twice then ...

leocardao: Done.

src/e3/anod/context.py
leocardao: Done.

src/e3/anod/context.py
```diff
@@ -439,7 +439,22 @@ def add_dep(spec_instance: Anod, dep: Dependency, dep_instance: Anod) -> None:
...
                 )
-            spec_instance.deps[dep.local_name] = dep_instance
+            if dep.kind != "source":
+                spec_instance.deps[dep.local_name] = dep_instance
+            else:
+                srcbuild_list = dep_instance.source_pkg_build
+                spec_instance.source_list[dep.local_name] = set()
+                if not srcbuild_list:
+                    raise AnodError(
+                        f"{dep.local_name} is marked as source_pkg but not sources are "
```
grouigrokon, line 449: **no** sources are defined ?

src/e3/anod/driver.py
```diff
@@ -61,7 +61,28 @@ def activate(self, sandbox: SandBox, spec_repository: AnodSpecRepository) -> Non
...
+                    self.anod_instance.deps[e.local_name] = dep_instance
+                else:
+                    srcbuild_list = dep_instance.source_pkg_build
+                    self.anod_instance.source_list[e.local_name] = set()
+                    if not srcbuild_list:
+                        raise AnodError(
+                            f"{e.local_name} is marked as source_pkg but not sources "
```
grouigrokon, line 71: Same here.

grouigrokon (commented):
Feel like you want with the two typoes. If you want to let them be, it is not an enormous matter :)

grouigrokon (approved):

Nikokrock (commented):
I think a cannot approve as some specs use Source dep to access spec functions. I saw them while rewriting some .anod. We need to change those specs first

Issue #609, opened 2023-07-24
Name generator - improve error tracing


JulienBortolussiAda:
It is currently difficult to know where an error come from for now.

For example if the spec A use the spec B with bad qualifiers then the spec B will complain. However, the culprit is the spec A.

PR #611, opened 2023-07-24
Add qualifier error tracing

JulienBortolussiAda:
Enhance the errors raised by the qualifier manager to show the parent (above in the dependency tree) spec.

### DIFF
--- a/src/e3/anod/qualifiers_manager.py
+++ b/src/e3/anod/qualifiers_manager.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, cast
 from hashlib import sha1
 
-from e3.anod.error import AnodError
+from e3.anod.error import AnodError, QualifierError
 
 import string
 
@@ -67,6 +67,10 @@ class QualifiersManager:
     computed using all the declared qualifiers and their values (at runtime).
     This allow the build_space names to be different for each different builds
     (assuming that different build <=> different set of qualifier values).
+
+    The QualifiersManager can raise:
+     * AnodError: when the error is related to the usage of the class
+     * QualifierError: when the error is related to the qualifiers parsed by the class
     """
 
     # Declare the qualifier types
@@ -399,7 +403,7 @@ class QualifiersManager:
                 )
             else:
                 raise AnodError(
-                    "An expected qualifier type was encountered during parsing "
+                    "An unexpected qualifier type was encountered during parsing "
                     f'Got "{qualifier["type"]}"'
                 )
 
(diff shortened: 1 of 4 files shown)


JulienBortolussiAda:
Solve https://github.com/AdaCore/e3-core/issues/609

grouigrokon:
You should run `pre-commit run --all-files` before pushing ...

tests/tests_e3/anod/test_qualifier_manager.py
```diff
@@ -171,13 +172,20 @@ class AnodDummy(Anod):
     with pytest.raises(AnodError) as err:
```
leocardao, line 172: Now it should raise a QualifierError right ?

JulienBortolussiAda: I have updated this every where I found it

src/e3/anod/qualifiers_manager.py
```diff
@@ -695,7 +696,7 @@ def __generate_qualifier_part(
                 return "_" + qualifier_name + "-" + qualifier_value
         else:
             raise AnodError(
```
grouigrokon, line 698: All those are qualifier errors I guess (Thanks Leo for pinpointing it). If so, raise a `QualifierError` and update the tests too :)
To me, all error message mentionning that an unexpected qualifier was encountered should raise a `QualifierError`.

JulienBortolussiAda: I forgot to add a comment to describe what is QualifierError.

Let's discuss it here :)

For me there are two kind of error raised by the QualifiersManager:
-  AnodError: The error is static (structural) and come from a mistake in the spec itself
-  QualifierError: The error is dynamic, it lies in the qualifiers passed to the spec, not in the spec itself. The culprit is the caller.

That is why I am only adding the context information to the QualifierErrors

What do you  think ?

JulienBortolussiAda: The one you are pointing at is fishy. To produce it, you have to mess up with the inner data structures of the QualifiersManager so for me it is an AnodError

grouigrokon: That's ok, I have read the various uses of the `__error` call or the `AnodError`. I hav no objection on you point of view :)
Now, a little comment in the `AnodError` case to say why we do not raise `QualifierError` could be good, but still, it's only my point of view.

JulienBortolussiAda: I have added a Little comment in the QualifiersManager class. That is what you meant ?

grouigrokon: Ok

leocardao (approved):
LGTM

Issue #607, opened 2023-07-24
Does not use the name generator in source mode


JulienBortolussiAda:
The qualifiers are not passed to the specs in source mode so the name generator cannot be used.

PR #612, opened 2023-07-24
Make the qualifiers manager errors clearer

JulienBortolussiAda:
Add the name of the spec in the errors messages to help finding out where the errors come from.

### DIFF
--- a/src/e3/anod/qualifiers_manager.py
+++ b/src/e3/anod/qualifiers_manager.py
@@ -743,7 +743,7 @@ class QualifiersManager:
     def __error(self, msg: str) -> None:
         """Raise an error and print the helper."""
         print(self.__get_helper())
-        raise AnodError(msg)
+        raise AnodError(self.anod_instance.name + ": " + msg)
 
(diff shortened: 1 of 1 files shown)


JulienBortolussiAda:
Solve https://github.com/AdaCore/e3-core/issues/607

src/e3/anod/qualifiers_manager.py
```diff
@@ -743,7 +743,7 @@ def build_space_name(self) -> str:
     def __error(self, msg: str) -> None:
         """Raise an error and print the helper."""
         print(self.__get_helper())
-        raise AnodError(msg)
+        raise AnodError(self.anod_instance.name + ": " + msg)
```
grouigrokon, line 746: Use of an f-string or (my favorite) call to `.format()` would prevent strings arithmetics.
Something like:

- `f"{self.anod_instance.name}: {msg}"`
- or `"{0}: {1}".format(self.anod_instance.name, msg)`

I guess that a dedicated test (or modification of an existing test) would be good too.

PR #613, opened 2023-07-24
Do not use the name generator in source mode

JulienBortolussiAda:
The qualifiers are always ignored by source packaging so it is useless.

### DIFF
--- a/src/e3/anod/spec.py
+++ b/src/e3/anod/spec.py
@@ -244,7 +244,7 @@ class Anod:
 
         # Create the QualifiersManager.
         # Skip if the name generator is disabled.
-        if self.enable_name_generator:
+        if self.enable_name_generator and self.kind != "source":
             self.qualifiers_manager = QualifiersManager(self)
             self.declare_qualifiers_and_components(self.qualifiers_manager)
             self.qualifiers_manager.parse(self.parsed_qualifier)
@@ -322,7 +322,7 @@ class Anod:
             build_space name otherwise.
         :rtype: str | None
         """
-        if self.enable_name_generator:
+        if self.enable_name_generator and self.kind != "source":
             return self.qualifiers_manager.build_space_name
         else:
             return self.name
--- a/tests/tests_e3/anod/test_qualifier_manager.py
+++ b/tests/tests_e3/anod/test_qualifier_manager.py
@@ -417,3 +417,13 @@ def test_qualifiers_manager():
     qualifiers_manager = QualifiersManager(Anod("", kind="build"))
     qualifiers_manager.parse({})
     qualifiers_manager.parse({})
+
+
+def test_qualifiers_with_source_primitive():
+    class SimpleAnod(Anod):
+        enable_name_generator = True
+        base_name = "simple"
+        name = "simple_anod"
+
+    simple_anod = SimpleAnod(kind="source", qualifier="")
+    assert simple_anod.build_space_name == "simple_anod"


JulienBortolussiAda:
Solve https://github.com/AdaCore/e3-core/issues/607

src/e3/anod/spec.py
```diff
@@ -245,9 +245,10 @@ def __init__(
         # Create the QualifiersManager.
         # Skip if the name generator is disabled.
         if self.enable_name_generator:
-            self.qualifiers_manager = QualifiersManager(self)
-            self.declare_qualifiers_and_components(self.qualifiers_manager)
-            self.qualifiers_manager.parse(self.parsed_qualifier)
+            if self.kind != "source":
```
grouigrokon, line 248: I know it is not the point of this MR, but knowing the number of references to the `"source"` string, I guess we could define an `is_source` field (which would refer to the `SOURCE_PRIMITIVE` value ?
As it is not the place, I will create a new Issue in github :) Sorry for the noise :)

src/e3/anod/spec.py
```diff
@@ -245,9 +245,10 @@ def __init__(
         # Create the QualifiersManager.
         # Skip if the name generator is disabled.
         if self.enable_name_generator:
```
leocardao, line 247: I think you should use `if self.enable_name_generator and self.kind != "source"`

src/e3/anod/spec.py
```diff
@@ -315,7 +316,10 @@ def build_space_name(self) -> str:
         :rtype: str | None
         """
         if self.enable_name_generator:
```
leocardao, line 318: You can rewrite this if like this: `return self.qualifiers_manager.build_space_name if self.enable_name_generator and self.kind != "source" else self.name`

src/e3/anod/spec.py
JulienBortolussiAda: Done

src/e3/anod/spec.py
JulienBortolussiAda: Done

leocardao:
LGTM

leocardao (approved):

src/e3/anod/spec.py
```diff
@@ -322,10 +322,11 @@ def build_space_name(self) -> str:
             build_space name otherwise.
         :rtype: str | None
         """
-        if self.enable_name_generator:
-            return self.qualifiers_manager.build_space_name
-        else:
-            return self.name
+        return (
```
Nikokrock, line 325: just use a if block to make code more readable

tests/tests_e3/anod/test_qualifier_manager.py
```diff
@@ -417,3 +417,13 @@ def declare_qualifiers_and_components(self, qualifiers_manager):
     qualifiers_manager = QualifiersManager(Anod("", kind="build"))
     qualifiers_manager.parse({})
     qualifiers_manager.parse({})
+
+
+def test_qualifiers_manager_and_spec():
```
Nikokrock, line 422: test_qualifiers_with_source_primitive ? or something similar ?

JulienBortolussiAda: Done

src/e3/anod/spec.py
JulienBortolussiAda: Done

PR #616, opened 2023-08-01
Use pyproject.toml instead of old setup.py

leocardao:
Remove src/e3/__init__.py to use:
    PEP 420 – Implicit Namespace Packages
    See: https://peps.python.org/pep-0420/

Update tox.ini to run into an isolated build.

Update mypy.ini and .pre-commit-config.yaml to work with the changes.


grouigrokon:
As long as the checks are not ok, I do not think I will spend time reviewing :)

PR #618, opened 2023-08-03
Update build space generator to support target info

Nikokrock:


src/e3/anod/qualifiers_manager.py
```diff
@@ -594,6 +639,21 @@ def compute_build_space_name(self) -> None:
...
+            # different build space each time
+            os_version = self.anod_instance.env.target.os.version
+            if os_version and os_version != "unknown":
+                bs.append(self.os_version_aliases.get(os_version, os_version))
+            # We also can test on different machines, so add it as well
+            machine = self.anod_instance.env.target.machine
+            if machine and machine != "unknown":
+                # and make sure the machine name is compatible with
+                # file paths: only keep letters
+                machine_str = "".join([c for c in machine if c.isalpha()])
```
grouigrokon, line 654: Why not `isalnum()` ?

leocardao: For me, it's neither isalnum nor isalpha.

A machine can be any string, it can have dash inside, for example.

Nikokrock: I think the logic was taken from existing code

Nikokrock: Maybe we can indeed accept more characters

Nikokrock: Will adjust the change to accept numbers and -

Nikokrock: Solved

liaiss (approved):

PR #619, opened 2023-08-03
Add + sign as valid qualifier name character

Nikokrock:
Needed by gprbuild

### DIFF
--- a/src/e3/anod/qualifiers_manager.py
+++ b/src/e3/anod/qualifiers_manager.py
@@ -11,7 +11,7 @@ if TYPE_CHECKING:
     from e3.anod.spec import Anod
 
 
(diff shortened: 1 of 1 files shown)


src/e3/anod/qualifiers_manager.py
```diff
@@ -11,7 +11,7 @@
     from e3.anod.spec import Anod
 
 
-VALID_NAME = re.compile(r"^[a-zA-Z0-9_.-]+$")
+VALID_NAME = re.compile(r"^[a-zA-Z0-9_.+-]+$")
```
grouigrokon, line 14: `[\w.+-]` ;)

grouigrokon: Who cares ;)

PR #625, opened 2023-08-30
e3-pypi-closure: Add option to allow pre-release version usage

leocardao:


src/e3/python/pypi.py
```diff
@@ -394,6 +397,9 @@ def __init__(
         self.load_cache_file()
         self.requirements: set[Requirement] = set()
         self.explicit_requirements: set[Requirement] = set()
+        self.allowed_prerelease = (
```
grouigrokon, line 400: I have to say that I like using `or`. Here, it would give:
```python
self.allowed_prerelease = allowed_prerelease or []
```

grouigrokon (approved):
Seems ok for me.

PR #626, opened 2023-09-04
Add .readthedocs.yaml (V2) configuration file

leocardao:
We do not support the latest versions of sphinx (at least sphinx v7.2.5).

Currently, we have forced readthedoc to use the latest version of sphinx
known to work with e3-core: v5.3.0.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,19 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+version: 2
+
+# Set the OS, Python version and other tools you might need.
+# OS can only be: ubuntu-20.04, ubuntu-22.04.
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+sphinx:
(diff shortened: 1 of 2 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,19 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+version: 2
+
+# Set the OS, Python version and other tools you might need.
+# OS can only be: ubuntu-20.04, ubuntu-22.04
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.9"
```
enzbang, line 10: I think you can use a more recent version, e.g. 3.11

leocardao: Done

leocardao:
> Do we have a way to test that ? If I get it ok, it is a readthedocs service which builds our doc. Build the doc with sphinx, or through tox seems ok (even if it has some errors).

you can see a "CI" here: https://readthedocs.org/projects/e3-core/builds/

enzbang (approved):

PR #627, opened 2023-09-05
Handle star ending requirements in pypi add_constraint

pierretr:
Some packages like google-api-client can have requirements that discard all micro versions of a package for a given major and minor version. This is the case of the google-api-client package (v 2.97) which for example requires google-api-core!=2.1.*.

### DIFF
--- a/tests/tests_e3/python/main_test.py
+++ b/tests/tests_e3/python/main_test.py
@@ -3,9 +3,11 @@ from e3.python.wheel import Wheel
 from e3.os.process import Run
 from e3.sys import python_script
 from e3.fs import mkdir
-from e3.python.pypi import PyPIClosure
+from e3.python.pypi import PyPIClosure, PyPIError
+
 import yaml
 import os
+import pytest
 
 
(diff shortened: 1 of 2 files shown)


tests/tests_e3/python/main_test.py
```diff
@@ -103,3 +105,46 @@ def test_pypi_closure_tool():
...
+    with PyPIClosure(
+        python3_version=10,
+        platforms=[
+            "x86_64-linux",
+        ],
+        cache_dir=".cache",
+        cache_file=".pypi.cache",
+    ) as pypi:
+        pypi.add_wheel(wheel1.path)
+        pypi.add_wheel(wheel3.path)
```
grouigrokon, line 148: I may be wrong, but I would say `wheel2`, then `wheel3`. Here, `src1` is added twice (which I guess is ok), but `src2` is not part of the wheel ?

pierretr: You're right. That's fixed.

PR #628, opened 2023-09-08
Update VCS ignore list

leocardao:


src/e3/fs.py
```diff
@@ -549,6 +549,9 @@ def splitall(path: str) -> tuple[str, ...]:
     ".gitreview",
     ".mailmap",
     ".idea",
+    ".python-version",
```
grouigrokon, line 552: What about `.flake8`. It does not contain any important data, but does it belong to the package ?

pierretr: `.github` could also be added to the list.

leocardao: @grouigrokon For .flake8 I don't think so, we provide the sources, for me that's part of it.

.github added.

Sorry for the delay, this PR was totally off my radar.

grouigrokon (approved):

PR #632, opened 2023-09-19
Authorize qualifier dictionaries in Dependency

JulienBortolussiAda:
Accept a qualifier dictionary in Dependency. This dictionary can be formatted the old fashion (some qualifier are not present) or following the "args" schemes (the falg qualifier value is True or False).

It will require a corresponding patch in the specs to update the wrappers of Dependency.


src/e3/anod/deps.py
```diff
@@ -79,8 +79,24 @@ def __init__(
         self.host = host
         self.target = target
         self.build = build
-        self.qualifier = qualifier
         self.local_name = local_name if local_name is not None else name
+
+        self.qualifier: str | None
+        if isinstance(qualifier, dict):
+            for q in set(qualifier.keys()):
```
Nikokrock, line 86: no need to cast on set

src/e3/anod/deps.py
```diff
@@ -79,8 +79,24 @@ def __init__(
...
+                v = qualifier[q]
```
Nikokrock, line 87: for q, v in qualifier.items()

src/e3/anod/deps.py
```diff
@@ -79,8 +79,24 @@ def __init__(
...
+                if isinstance(v, bool):
+                    # It's a tag qualifier
+                    if v:
+                        qualifier[q] = ""
+                    else:
+                        qualifier.pop(q)
+            self.qualifier = ",".join(
```
Nikokrock, line 94: sort by key and ensure that list, set or frozenset are ordered too

src/e3/anod/deps.py
JulienBortolussiAda: I cannot do that because I am changing the size of the dictionary in the loop (poping an element) and python doesn't like that.

src/e3/anod/deps.py
JulienBortolussiAda: The cast to set is way to copy it. Otherwise python complains because the size of the dictionary change during the iteration.

src/e3/anod/deps.py
JulienBortolussiAda: Done

src/e3/anod/deps.py
grouigrokon: To me it's ok. Using a `set` or a `list` seems to do the same.

```python
>>> a = {1: "one", 2: "two", 3: "three"}
>>> for k in list(a.keys()):
...     a[1 + len(a)] = "something"
...     print(k)
... 
1
2
3
>>> a
{1: 'one', 2: 'two', 3: 'three', 4: 'something', 5: 'something', 6: 'something'}
>>> for k in a:
...     a[1 + len(a)] = "something"
...     print(k)
... 
1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
>>>
````

grouigrokon (approved):
As long as Nico does not have any other comment, and that your answers are right, I think we can approve the MR.

Issue #630, opened 2023-09-13
Add a new kind of qualifier: key_set qualifiers


JulienBortolussiAda:
Some qualifiers are actually expecting a list of value.

The current work around is to use key_value qualifier without choices and then to the parsing in the spec.

key_set qualifier will handle semi-column (";") separated list of values. The value will appears ordered in the build_space name.

grouigrokon:
BTW, ";" is semi-column. "," is comma :)

PR #633, opened 2023-09-19
Add support for key_set qualifier

JulienBortolussiAda:
key_set qualifiers expect semi-colon separated list as value.

#630

### DIFF
--- a/src/e3/anod/qualifiers_manager.py
+++ b/src/e3/anod/qualifiers_manager.py
@@ -68,7 +68,7 @@ class QualifierDeclaration(metaclass=abc.ABCMeta):
         )
 
     @property
-    def default(self) -> str | bool | None:
+    def default(self) -> str | bool | frozenset[str] | None:
         """Return default value for qualifier.
 
         :return: if None is returned it means the qualifier has not
@@ -77,11 +77,13 @@ class QualifierDeclaration(metaclass=abc.ABCMeta):
         return None  # all: no cover
 
     @abc.abstractmethod
-    def value(self, value: str | bool) -> str | bool:
-        """Validate qualifier value and return the effective one."""
+    def value(self, value: str) -> str | bool | frozenset[str]:
+        """Compute the value of qualifier given the use input."""
 
     @abc.abstractmethod
-    def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
+    def repr(
+        self, value: str | bool | frozenset[str], hash_pool: list[str] | None
+    ) -> str:
         """Compute a string representation of a qualifier.
 
         :param value: the effective value associated with the qualifier
@@ -140,12 +142,14 @@ class KeyValueDeclaration(QualifierDeclaration):
         self._default = default
 
     @property
-    def default(self) -> str | bool | None:
+    def default(self) -> str | bool | frozenset[str] | None:
         """See QualifierDeclaration.default."""
         return self._default
 
-    def value(self, value: str | bool) -> str | bool:
+    def value(self, value: str) -> str | bool | frozenset[str]:
         """See QualifierDeclaration.value."""
+        if not isinstance(value, str):
+            raise AnodError("Key value qualifiers can only parse a string value.")
         if self.choices is not None and value not in self.choices:
             choices_str = ", ".join((f"'{choice}'" for choice in self.choices))
             raise AnodError(
@@ -154,10 +158,13 @@ class KeyValueDeclaration(QualifierDeclaration):
             )
         return value
 
-    def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
+    def repr(
+        self, value: str | bool | frozenset[str], hash_pool: list[str] | None
+    ) -> str:
         """See QualifierDeclaration.repr."""
         if not value:
-            # An empty value for tag value should lead to an empty representation
+            # An empty value for a key_value qualifier should lead to an empty
+            # representation
             str_repr = ""
         elif (
             value == self.default
@@ -173,8 +180,9 @@ class KeyValueDeclaration(QualifierDeclaration):
             list_repr = []
             if not self.repr_omit_key:
                 list_repr.append(self.repr_name)
-            if value:
-                list_repr.append(str(value))
+
+            # value as been checked early on
+            list_repr.append(str(value))
 
             # And join them with a dash.
             str_repr = "-".join(list_repr)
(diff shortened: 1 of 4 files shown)


JulienBortolussiAda:
Maybe I could add a way to set the separator

src/e3/anod/qualifiers_manager.py
```diff
@@ -219,6 +227,122 @@ def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
...
+            description=description,
+            repr_in_hash=repr_in_hash,
+            repr_name=repr_name,
+        )
+        self.repr_omit_key = repr_omit_key
+
+        # Check if the default is valid
+        default_values: frozenset[str] | None = None
+        if default is not None:
+            default_values = frozenset(default.split(";"))
```
liaiss, line 266: Define a constant for the ; separator, it should make things more understandable

src/e3/anod/qualifiers_manager.py
```diff
@@ -219,6 +227,122 @@ def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
...
+
+        return value_set
+
+    def repr(
+        self, value: str | bool | frozenset[str], hash_pool: list[str] | None
+    ) -> str:
+        """See QualifierDeclaration.repr."""
+        assert isinstance(value, frozenset)
+        if not value:
+            # An empty value for tag value should lead to an empty representation
```
liaiss, line 317: it seems that the comment is not corresponding the key_set case here

src/e3/anod/qualifiers_manager.py
```diff
@@ -436,6 +564,63 @@ def declare_key_value_qualifier(
...
+        if not test_only or self.anod_instance.kind == "test":
+            self.qualifier_decls[name] = KeySetDeclaration(
+                origin=self.origin,
+                name=name,
+                description=description,
+                repr_name=repr_alias,
+                repr_in_hash=repr_in_hash,
+                default=default,
+                choices=choices,
+                repr_omit_key=repr_omit_key,
```
liaiss, line 621: I don't understand the condition, here. What I read not test_only means all the case or test, so I don't understand why you do this check?

src/e3/anod/qualifiers_manager.py
JulienBortolussiAda: Done

src/e3/anod/qualifiers_manager.py
```diff
@@ -219,6 +228,125 @@ def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
...
+        super().__init__(
+            origin=origin,
+            name=name,
+            description=description,
+            repr_in_hash=repr_in_hash,
+            repr_name=repr_name,
+        )
+        self.repr_omit_key = repr_omit_key
+
+        self.separator = ";"
```
liaiss, line 264: Even if an attribute is better, we should define a constant for ";", for example SEMI_COLON_SEPARATOR

JulienBortolussiAda: At the class level then ? I doesn't think it make any sens outside of it.

I went for an instance attribute so that we can choose to let the user select his separator later on.

src/e3/anod/qualifiers_manager.py
```diff
@@ -140,12 +142,13 @@ def __init__(
...
     @property
-    def default(self) -> str | bool | None:
+    def default(self) -> str | None:
         """See QualifierDeclaration.default."""
         return self._default
 
-    def value(self, value: str | bool) -> str | bool:
+    def value(self, value: str | bool | frozenset[str]) -> str:
         """See QualifierDeclaration.value."""
+        assert isinstance(value, str)
```
Nikokrock, line 151: Raise an Anod Error rather than an assert

src/e3/anod/qualifiers_manager.py
```diff
@@ -140,12 +142,13 @@ def __init__(
         self._default = default
 
     @property
-    def default(self) -> str | bool | None:
+    def default(self) -> str | None:
```
Nikokrock, line 145: Keep the same signature as the parent class

src/e3/anod/qualifiers_manager.py
```diff
@@ -219,6 +228,125 @@ def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
...
+class KeySetDeclaration(QualifierDeclaration):
+    def __init__(
+        self,
+        origin: str,
+        name: str,
+        description: str,
+        repr_in_hash: bool = False,
+        repr_name: str | None = None,
+        repr_omit_key: bool = False,
+        default: str | None = None,
```
Nikokrock, line 240: default should be a set

src/e3/anod/qualifiers_manager.py
Nikokrock: Move the declaration at class level. We don't want to support various possibility

src/e3/anod/qualifiers_manager.py
```diff
@@ -219,6 +228,125 @@ def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
...
+                    raise AnodError(
+                        f"{self.origin}: In '{self.name}', default value(s) "
+                        f"({wrong_values_str}) should be in ({choices_str})"
+                    )
+
+        self.choices = choices
+        self._default = default_values
+
+    @property
+    def default(self) -> frozenset[str] | None:
```
Nikokrock, line 287: Use same signature as root class

src/e3/anod/qualifiers_manager.py
Nikokrock: I think it's ok

src/e3/anod/qualifiers_manager.py
```diff
@@ -219,6 +229,118 @@ def repr(self, value: str | bool, hash_pool: list[str] | None) -> str:
...
+        self._default: frozenset[str] | None = (
+            frozenset(default) if default is not None else None
+        )
+
+    @property
+    def default(self) -> str | bool | frozenset[str] | None:
+        """See QualifierDeclaration.default."""
+        return self._default
+
+    def value(self, value: str | bool | frozenset[str]) -> str | bool | frozenset[str]:
```
Nikokrock, line 291: there is no case in which a frozen set is accepted as input

JulienBortolussiAda: I have also remove bool for the same reason. If I missed something, let me know.

Nikokrock (approved):

PR #634, opened 2023-09-20
e3.diff.patch raise EmptyDiffError when doing nothing

JulienBortolussiAda:
Raise a specialize error when e3.diff.patch do nothing. These can happen when all the patched file are discarded.

### DIFF
--- a/src/e3/diff.py
+++ b/src/e3/diff.py
@@ -16,6 +16,17 @@ logger = e3.log.getLogger("diff")
 
 
 class DiffError(e3.error.E3Error):
+    """Error returned when the patch command fails."""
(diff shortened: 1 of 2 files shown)


src/e3/diff.py
```diff
@@ -19,6 +19,10 @@ class DiffError(e3.error.E3Error):
     pass
 
 
+class EmptyDiffError(DiffError):
```
Nikokrock, line 22: Add a docstring

JulienBortolussiAda: Done

liaiss (approved):

PR #636, opened 2023-09-22
Use Specifier class to check version constraints

adanaja:
Package.add_constraint implements its own way to check for version constraints, while packaging.specifiers.Specifier can be used


grouigrokon:
The `packaging` module is external to `python`. Would the `requirements.txt` (or `setup.py` here) need an update ?

PR #639, opened 2023-10-04
Fix doc build failure

leocardao:
Last time, the doc build failed because rtd was not fully supported by Sphinx7. (See: [https://github.com/readthedocs/sphinx_rtd_theme/issues/1463](https://github.com/readthedocs/sphinx_rtd_theme/issues/1463))

This problem has now been corrected, so we want to use the latest rtd-compatible version of Sphinx.

Also fix the requirements version used in order to avoid future build failure.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
-sphinx==5.3.0
-sphinx-autoapi
+sphinx==7.2.6
+sphinx-autoapi==3.0.0
+sphinx_rtd_theme==1.3.0


docs/requirements.txt
```diff
@@ -1,2 +1,2 @@
-sphinx==5.3.0
+sphinx
 sphinx-autoapi
```
grouigrokon: No need to add `sphinx_rtd_theme` ?

grouigrokon (approved):

docs/requirements.txt
leocardao: I wanted to see if it worked without it, but since it didn't, I added it. :)

enzbang: RTD injects the needed packages apparently:

python -m pip install --upgrade --no-cache-dir sphinx readthedocs-sphinx-ext

PR #640, opened 2023-10-05
Add readme management into anod spec

leocardao:


src/e3/anod/spec.py
```diff
@@ -274,6 +274,25 @@ def enable_name_generator(self) -> bool:
         """
         return False
 
+    @property
+    def readme_filename(self) -> str:
+        """The readme filename to use.
+
+        :return: The readme filename to use.
+            Note that the return value cannot be None or an empty string
```
grouigrokon, line 282: You may want to use the ``.. note::`` RST markup:

```
:return: The readme filename to use.

.. note:: The return value cannot be None or an empty string.
```

src/e3/anod/spec.py
```diff
@@ -274,6 +274,25 @@ def enable_name_generator(self) -> bool:
...
+        """
+        return f"README-{self.component}"
+
+    @property
+    def readme_ext(self) -> str:
+        """The readme extension to use.
+
+        :return: The extension used by the readme file.
+            Note that if this return value begin by dot(s), these dots will be
```
grouigrokon, line 291: Same for the `.. note::`

src/e3/anod/spec.py
```diff
@@ -274,6 +274,25 @@ def enable_name_generator(self) -> bool:
...
+            discarded (Example: ..txt = .txt = txt)
```
grouigrokon, line 292: Seeing the method body, I would say that the documentation is wrong :)
Either say _the returned value should be stripped of its starting dots._, or change how the property is used.

leocardao: Currently, this comment isn't true, but it should become true soon. And it's almost certain that I'll forget to update this docstring, so I did it a little early 😄 

Do you think this is really a problem?

src/e3/anod/spec.py
leocardao: Done.

src/e3/anod/spec.py
leocardao: Done.

src/e3/anod/spec.py
leocardao: We've completely changed the way the code works, so this comment is now obsolete. I'll take the liberty of resolving it.

src/e3/anod/spec.py
```diff
@@ -274,6 +274,17 @@ def enable_name_generator(self) -> bool:
         """
         return False
 
+    @property
+    def readme(self) -> tuple[str, str] | None:
```
liaiss, line 278: for me, readme_info is clearer because here the tuple return to information not linked:
  - the full path of the readme file (source)
  - the destination name of the readme in the store.

liaiss (approved):

Issue #641, opened 2023-10-06
Bump to python 3.11 and 3.12


JulienBortolussiAda:
In the context of our annual bump to newer python versions, we need to move forward to support python 3.11 and 3.12 instead of 3.9 and 3.10

PR #642, opened 2023-10-06
Add testing for python 3.11

JulienBortolussiAda:
Set up the test for python 3.11 and 3.12 in the CI

#641


grouigrokon:
Shall the `tox` environment be updated too ?

liaiss (approved):

JulienBortolussiAda:
@grouigrokon, Yes, we will have to do that too but I don't think it is a good idea yet. If we bump it to python 3.10, all the devs will need to have python 3.10 on their computer while the one we support (and everyone has) in e3-distrib is python 3.9

PR #645, opened 2023-10-17
Slsa provenance

grouigrokon:
Added SLSA provenance package.

This provenance package is the implementation of the SLSA provenance
specifications at https://slsa.dev/spec/v1.0/provenance

For now, at least two enhancements are possible:

- Some parameters are TypeURI or ResourceURI, but the specification
  says they are strings. Setting those attributes as `str` and having
  a check that they fit with TypeURI or ResourceURI classes could be
  better.

Documentation fixes
-------------------
Removed some warning about autoapi documentation, and some about
possible classes mixes.

Still the warning about PIPE import in TYPE_CHECKING could not be resolved.

Review changes
--------------
- Removed DigestSet and DigestAlgorithm enum. A digest set is considered
  a dictionary of str: str. The ResourceDescriptor's digest may be set,
  or new digest added by the add_digest() method.
- The dir_hash() method has been moved to ResourceDescriptor ... maybe not
  the best place though.
- Fixed a minor typo in the dsse package documentation
- Added `as_dict()`, `as_json()`, `load_dict()` and `load_json()` for all
  the objects.
- Added a provenance file example in the tests, and load it.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -41,6 +41,9 @@ autoapi_dirs = ["../../src/e3"]
 templates_path = ["_templates"]
 autoapi_template_dir = "source/autoapi_templates"
 
+# Remove warnings for auto API template.
+exclude_patterns = ["autoapi_templates/index.rst"]
+
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
 #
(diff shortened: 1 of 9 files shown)


grouigrokon:
Comments from review have been taken into account.

Nikokrock (approved):

PR #647, opened 2023-10-19
Fix for rlimit foreground test.

grouigrokon:
Using the `TERM=dummy` environment fixed the prompt coloring issue. Some minor changes to the test file itself and `setup.py` to avoid IDE warnings.


Nikokrock (commented):
One minor comment

Nikokrock (approved):

PR #650, opened 2023-10-27
Modify Trie code to avoid a mypy error

enzbang:
cursor is a dict, cursor.get(.., None) returns another dict or None mypy is being confused as the variable is overwritten, use another temp variable to handle the case where letter is not in the cursor dict


grouigrokon:
LGTM

Nikokrock (approved):

PR #652, opened 2023-11-15
Add request-cache package dependency to e3-core

enzbang:


grouigrokon (approved):
Seems like using `requests_cache` also works. Pypi's magic ...

PR #656, opened 2023-11-21
Updated the Document.add_package() method

grouigrokon:
As seen with Nico, the `Document.add_package()` method has been updated (no need to be backward compatible from Nico's point of view) to use a `Package` for input parameter, rather than 14 parameters used to create a `Package` within the method.

- Modified the `Document.add_package()` method to take a package for input, not many parameters which are used to build a package in the end.
- Updated the tests for full coverage of the SPDX module, and to take into account the modifications of the `add_package()` method.
- Added more documentation for Package class fields.
- Added package comments to the package class.
- Added links to the specification for the package fields.

### DIFF
--- a/src/e3/spdx.py
+++ b/src/e3/spdx.py
@@ -301,7 +301,8 @@ class Tool(Entity):
 class PackageName(SPDXEntryStr):
     """Identify the full name of the package.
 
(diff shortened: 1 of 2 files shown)


leocardao (approved):
LGTM

PR #661, opened 2023-12-18
diff.py(diff): make number of context lines configurable

glacambre:


glacambre:
@enzbang I don't know who would be the right reviewer for this and you're the last person who touched this code. Could you have a look, or assign the right reviewer? (I don't have enough permissions to assign reviewers.)

src/e3/diff.py
```diff
@@ -92,7 +94,9 @@ def diff(
                 line for line in contents[k] if re.search(ignore, line) is None
             ]
 
-    return "".join(unified_diff(contents[0], contents[1], item1name, item2name, n=1))
+    return "".join(
+        unified_diff(contents[0], contents[1], item1name, item2name, n=context)
```
grouigrokon, line 98: Just a little remark. If we use named variable for `n`, we could also name the other name with default values (*item1name* and *item2name* are `fromfile` and `tofile` if I am not wrong).
Still this is just a remark.

glacambre: The reason `n` is named is that it can't be positional, as the code skips over `fromfiledate` and `tofiledate` - but the other arguments can be positional. I guess we should refer to the AdaCore Python Style Guide (is there such a thing? I found one for Ada but not Python...) to know what we should do.

grouigrokon: Nah, as said, it is just a remark. I also tend to use names everywhere :) I just wanted to have @glacambre advice on that :)

grouigrokon (approved):

leocardao:
Apart from @grouigrokon 's comment, my only point concerns the type of `context`. It's supposed to be an unsigned int (this type doesn't exist in python).

I haven't yet found a way of specifying this type that works with mypy and doesn't require more code (the `Annotated` type does exist, but I've tested it and our mypy didn't detect an error).

Also, so that it's clear to everyone, can you at least add in the comment that it's a positive int?

Maybe also add an `assert context >= 0` or an `if context <= 0: raise ValueError` (I don't really know if this will help mypy understand what's going on, but at least it's explicit 😄)

src/e3/diff.py
```diff
@@ -37,6 +37,7 @@ def diff(
     item1name: str = "expected",
     item2name: str = "output",
     ignore_white_chars: bool = True,
+    context: int = 1,
```
leocardao, line 40: Apart from @grouigrokon 's comment, my only point concerns the type of context. It's supposed to be an unsigned int (this type doesn't exist in python).

I haven't yet found a way of specifying this type that works with mypy and doesn't require more code (the `Annotated` type does exist, but I've tested it and our mypy didn't detect an error).

Also, so that it's clear to everyone, can you at least add in the comment that it's a positive int?

Maybe also add an `assert context >= 0` or an `if context <= 0: raise ValueError` (I don't really know if this will help mypy understand what's going on, but at least it's explicit 😄)

leocardao (changes requested):

src/e3/diff.py
glacambre: Done.

I was about to suggest enforce this style rule with automation, but it seems that despite the plethora of python linters/static analyzers, there's no pre-existing rule that checks for this. It seems implementabale with CodeQL though.

src/e3/diff.py
glacambre: Done.

leocardao (approved):

pmderodat:
Hello! Would it be possible to publish a new e3-core release so that e3-testsuite can use this new feature? Thanks!

liaiss:
> Hello! Would it be possible to publish a new e3-core release so that e3-testsuite can use this new feature? Thanks!

you means in the package registry, right?

pmderodat:
Yes, on Pypi.

PR #666, opened 2024-01-17
Fix the SLSA build metadata tests

grouigrokon:
A SLSA predicate object contains run details, which itself contains some build metadata (start and end times).

A `BuildMetadata` class is thus initialized with two timestamps. Those timestamps are check and transformed to UTC timezone timestamps.

Unfortunately the timezone transformation was erasing the timestamp itself.

By using `datetime.astimezone()` method (and not `datetime.utcnow()` anymore), the test failure is fixed. To make sure this does not happen anymore, a time sleep has been added in some tests.

Fixes https://github.com/AdaCore/e3-core/issues/668

### DIFF
--- a/tests/tests_e3/slsa/provenance_test.py
+++ b/tests/tests_e3/slsa/provenance_test.py
@@ -9,6 +9,7 @@ import pytest
 from datetime import datetime, timezone
 from dateutil import parser as date_parser
 from pathlib import Path
+from time import sleep
 from typing import Any
 
 from e3.slsa.provenance import (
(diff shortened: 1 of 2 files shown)


tests/tests_e3/slsa/provenance_test.py
```diff
@@ -425,6 +427,7 @@ def test_predicate_load_json() -> None:
     bd: Predicate.BuildDefinition = create_valid_build_definition()[-1]
     rd: Predicate.RunDetails = create_valid_run_details()[-1]
     predicate: Predicate = Predicate(build_definition=bd, run_details=rd)
+    sleep(2.0)
```
leocardao, line 430: I think you should add a comment to explain why this sleep is important.

grouigrokon: I guess I have to create an issue then, so that I can refer to it here ...

grouigrokon: Done (PR description updated).

PR #667, opened 2024-01-17
e3-pypi-closure: Add yanked management

leocardao:
As stated in PEP_592, a yanked package should not be installed unless it is the only one matching a version specifier.

It used to be possible for e3-pypi-closure to choose a yanked package, if it was simply the most recent version of the package.

With this patch, no yanked package will be installed without explicit use of the --allow-yanked option.

Also add minor improvement to e3-pypi-closure --help.

### DIFF
--- a/src/e3/python/pypi.py
+++ b/src/e3/python/pypi.py
@@ -388,6 +388,10 @@ class PyPIClosure:
             data is cached a maximum of 24h. The cache contains results of requests to
             PyPI.
         :param pypi_url: set Python package registry URL. Default is PyPI
+        :param allowed_prerelease: list of package names authorized to be into
+            pre-release.
+        :param allowed_yanked: list of package names authorized to have yanked flags set
+            to true (see: PEP_592).
         """
         self.cache_file = cache_file
         self.cache_dir = cache_dir
--- a/src/e3/python/pypi.py
+++ b/src/e3/python/pypi.py
@@ -67,6 +67,11 @@ class PackageFile:
         return self.data["url"]
 
     @property
+    def is_yanked(self) -> bool:
+        """Return whether the package is yanked."""
+        return self.data.get("yanked", False)
+
+    @property
     def is_wheel(self) -> bool:
         """Return whether the package is a wheel."""
         return self.kind == "bdist_wheel"
@@ -191,6 +196,7 @@ class Package:
         for version in self.data.get("releases", {}):
             try:
                 v = packaging.version.parse(version)
+
                 if (not v.is_prerelease and not v.is_devrelease) or (
                     v.is_prerelease and self.name in pypi.allowed_prerelease
                 ):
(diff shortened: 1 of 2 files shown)


src/e3/python/pypi.py
```diff
@@ -365,6 +377,7 @@ def __init__(
         cache_file: str | None = None,
         pypi_url: str = "https://pypi.org/pypi",
         allowed_prerelease: list[str] | None = None,
+        allowed_yanked: list[str] | None = None,
```
liaiss, line 380: You should add the corresponding DocString for this parameter, allowed_prerelease is also missing

src/e3/python/pypiscript.py
```diff
@@ -88,6 +88,14 @@ def main() -> None:
         default=None,
         help="Allow to use pre-release version for some requirements",
     )
+    m.argument_parser.add_argument(
+        "--allow-yanked",
+        dest="allowed_yanked",
+        metavar="REQUIREMENT",
```
grouigrokon, line 94: Use `"PACKAGE"` instead ?

src/e3/python/pypiscript.py
```diff
@@ -88,6 +88,14 @@ def main() -> None:
...
+        nargs="*",
```
grouigrokon, line 95: Use `nargs="+"` would be more appropriate I think.

src/e3/python/pypiscript.py
leocardao: Done.

grouigrokon:
No additional tests ?
Not sure how easy it is to create a yanked package, but I think that simply editing its metadata would do ?

liaiss (approved):

leocardao:
@grouigrokon I don't know if this is feasible, but in any case, the current code doesn't seem to allow it easily.

Currently, in the tests, we generate a setup.py for this, but I haven't found any information to mark a package "yanked" in this file.

So we'd have to create a fake pypi server. If you think it's worth it, I'll do it, but this PR is expected by other teams so I suggest you do it in another PR.

Do you agree?

leocardao:
> I agree that it would be a bit of work. Let's keep it for later. My idea was to:
> 
> * create a wheel
> * unzip
> * update the METADATA file to mark it as yanked
> * rezip it (wheel file are zip files)

Perfect, I'll open the issue and merge this PR. Thank you.

leocardao:
For the record: @grouigrokon your method doesn't seem to work. I just downloaded the setuptools-scm wheel version 8.0.0, which is a yanked package. There is unfortunately no trace of "yanked", in the metadata. I think it's pypi who says if something is yanked, not the wheels. It is therefore necessary to simulate pypi.

I put the issue with the label `good first issue` (even if I had planned to do it ) ... I think that ultimately it is not a good issue to enter the code.

grouigrokon:
Ok, this information seems to be coming from the pypi server :'(

PR #669, opened 2024-01-18
Add support for S3 bucket as source for anod

Nikokrock:
Part of it/org/software-supply-chain/product-security/issues#24

### DIFF
--- a/src/e3/anod/checkout.py
+++ b/src/e3/anod/checkout.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import hashlib
 import json
 import os
+import re
 import tempfile
 from contextlib import closing
 from typing import TYPE_CHECKING
@@ -19,6 +20,8 @@ if TYPE_CHECKING:
     from typing import Literal
     from collections.abc import Callable
     from e3.mypy import assert_never
+
+    SupportedVCS = Literal["git"] | Literal["svn"] | Literal["external"] | Literal["s3"]
 
 logger = e3.log.getLogger("e3.anod.checkout")
 
@@ -57,7 +60,7 @@ class CheckoutManager:
 
     def update(
         self,
-        vcs: Literal["git"] | Literal["svn"] | Literal["external"],
+        vcs: SupportedVCS,
         url: str,
         revision: str | None = None,
     ) -> ReturnValue:
@@ -84,6 +87,8 @@ class CheckoutManager:
             update = self.update_svn
         elif vcs == "external":
             update = self.update_external
+        elif vcs == "s3":
+            update = self.update_s3
         else:
             assert_never()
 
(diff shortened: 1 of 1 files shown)


src/e3/anod/checkout.py
```diff
@@ -20,6 +21,8 @@
     from collections.abc import Callable
     from e3.mypy import assert_never
 
+    SupportedVCS = Literal["git"] | Literal["svn"] | Literal["external"] | Literal["s3"]
```
grouigrokon, line 24: Would an `Enum` do it ? Use of `Literal` may be hard to cope with in some cases (type issue).

enzbang: That would be a backward incompatible change. Not sure this is worth it for just type-checking improvements.

Nikokrock: Agree here with olivier.

src/e3/anod/checkout.py
```diff
@@ -102,6 +107,50 @@ def update(
...
+        :param revision: ignored
+        """
+        if os.path.isdir(self.working_dir):
+            old_commit = get_filetree_state(self.working_dir)
+        else:
+            old_commit = ""
+
+        # Extract profile from the url and perform basic
+        # url validation
+        m = re.match("s3://([a-zA-Z0-9_-]+@)?(.+)$", url)
```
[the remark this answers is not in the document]
Nikokrock: This is not a URL strictly speaking just a URI but I can try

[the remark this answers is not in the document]
Nikokrock: Not in the list of supported scheme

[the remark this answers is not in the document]
Nikokrock: See https://docs.python.org/3/library/urllib.parse.html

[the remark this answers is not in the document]
grouigrokon:
```python
>>> from urllib.parse import urlparse
>>> url1 = "s3://profile@bucket/suffix"
>>> url2 = "s3://bucket/suffix"
>>> urlparse(url1)
ParseResult(scheme='s3', netloc='profile@bucket', path='/suffix', params='', query='', fragment='')
>>> urlparse(url2)
ParseResult(scheme='s3', netloc='bucket', path='/suffix', params='', query='', fragment='')
>>> 
```
Except for the `profile@` stuff it's not so bad. I suggested that just because the regular expression with authorized characters and so on, does not seem to be a good idea (to me).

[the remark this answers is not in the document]
Nikokrock: I have counter examples when you start using some weird character:

$ cat s3_keys_parsing
from urllib.parse import urlparse

print(urlparse("s3://bucket/some/prefix/objectwith&#13;carriagereturn"))

$ python ./s3_keys_parsing
ParseResult(scheme='s3', netloc='bucket', path='/some/prefix/objectwith&', params='', query='', fragment='13;carriagereturn')

You can use any UTF-8 character in an object key name. However, using certain characters in key names can cause problems with some applications and protocols. The following guidelines help you maximize compliance with DNS, web-safe characters, XML parsers, and other APIs.

enzbang (approved):

enzbang:
Note that there are some errors

```
src/e3/anod/checkout.py:24: error: Invalid type alias: expression is not a valid type  [valid-type]
src/e3/anod/checkout.py:24: error: Unsupported left operand type for | ("object")  [operator]
src/e3/anod/checkout.py:63: error: Variable "e3.anod.checkout.SupportedVCS" is not valid as a type  [valid-type]
src/e3/anod/checkout.py:63: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
src/e3/anod/checkout.py:93: error: Missing positional argument "value" in call to "assert_never"  [call-arg]
Found 4 errors in 1 file (checked 96 source files)
```

src/e3/anod/checkout.py
leocardao, line 24: You must use `typing.Union`. `|` is not supported for type aliasing.

src/e3/anod/checkout.py
```diff
@@ -84,6 +87,8 @@ def update(
             update = self.update_svn
         elif vcs == "external":
             update = self.update_external
+        elif vcs == "s3":
+            update = self.update_s3
         else:
             assert_never()
```
leocardao, line 93: This function takes one argument as parameter.

src/e3/anod/checkout.py
```diff
@@ -102,6 +107,50 @@ def update(
             )
         return result
 
+    def update_s3(
+        self, url: str, revision: str | None
+    ) -> tuple[ReturnValue, str, str]:  # all: no cover
+        """Update working dir using a S3 bucket as source.
+
+        :param url: s3 url to a s3 directory. s3://[profile@]bucket/suffix
+        :param revision: ignored
```
leocardao, line 116: Why create an ignored parameter?

src/e3/anod/checkout.py
```diff
@@ -102,6 +107,50 @@ def update(
...
+        if os.path.isdir(self.working_dir):
+            old_commit = get_filetree_state(self.working_dir)
+        else:
+            old_commit = ""
+
+        # Extract profile from the url and perform basic
+        # url validation
+        m = re.match("s3://([a-zA-Z0-9_-]+@)?(.+)$", url)
+        if m is None:
+            raise e3.error.E3Error("invalid s3 url: {url}")
```
leocardao, line 127: missing `f"..."`

src/e3/anod/checkout.py
leocardao, line 125: Maybe should we compile this regex?

leocardao (changes requested):

Issue #671, opened 2024-01-18
e3-pypi-closure: Add yanked management test


leocardao:
Currently, e3-pypi-closure has no tests for yanked packet management. As said in #667, we simply need to add some tests.

PR #674, opened 2024-01-24
Implement pypi simulator + e3-pypi-closure yanked tests

leocardao:
The pypi simulator is used to simulate the interactions between our code and the real pypi.

Addresses are mocked, and for the moment it only simulates requests in https://pypi.org/pypi/<YOUR PACKAGE>/json format.

Test json metadata must be downloaded in advance and placed in tests/tests_e3/pypi_data/json.

This PR also add e3-pypi-closure yanked tests.

Related to #671


tests/tests_e3/conftest.py
```diff
@@ -1,15 +1,26 @@
 # type: ignore
+from __future__ import annotations
+
 import os
+from os.path import abspath, dirname, join as path_join, isfile
```
grouigrokon, line 5: I do not know if the package importer is able to read that, but I am not ...\
Three imports for two names ?

tests/tests_e3/conftest.py
```diff
@@ -164,3 +175,87 @@ def pytest_sessionfinish(session, exitstatus):
...
+        with open(pkg.path, "rb") as f:
+            result = f.read()
+
+        context.status_code = 200
+        return result
+
+    def get_metadata(self, request: Any, context: Any) -> str:
+        m = self.MATCHER.match(request.url)
+        if not m:
+            raise ValueError("mocked pypi received an unexpected request")
```
grouigrokon, line 216: Quite surely adding the failing request URL in the error message could help.

tests/tests_e3/conftest.py
```diff
@@ -164,3 +175,87 @@ def pytest_sessionfinish(session, exitstatus):
...
+        package = m.group("package")
+        path = path_join(self.DATA_DIR, "json", f"{package}.json")
+        if not isfile(path):
+            context.status_code = 404
+            return {"message": "Not Found"}
```
grouigrokon, line 221: What about `f"Package {package}.json not found"` ?

tests/tests_e3/conftest.py
```diff
@@ -164,3 +175,87 @@ def pytest_sessionfinish(session, exitstatus):
...
+
+        try:
+            with open(path) as json_file:
+                result = json_loads(json_file.read())
+
+            if "releases" not in result:
+                raise Exception("Bad metadata")
+        except Exception:
```
grouigrokon, line 229: Here we lose the exception history. Maybe creating an `InvalidMetadataError` exception, and then using `except InvalidMetadataError` would help tracking it.

tests/tests_e3/conftest.py
```diff
@@ -164,3 +175,87 @@ def pytest_sessionfinish(session, exitstatus):
...
+
+class PypiSimulator:
+    PYPI_URL = "https://pypi.org"
+    MATCHER = regex_compile(f"{PYPI_URL}/pypi/(?P<package>.*)/json")
+    DATA_DIR = path_join(dirname(abspath(__file__)), "pypi_data")
+
+    def __init__(self, requests_mock: Any) -> None:
+        self.requests_mock = requests_mock
+        self.requests_mock.stop()
+        self.download_url_mocked = set()
```
grouigrokon, line 188: What about `mocked_download_urls` ?

tests/tests_e3/conftest.py
```diff
@@ -164,3 +175,87 @@ def pytest_sessionfinish(session, exitstatus):
...
+        except Exception:
+            context.status_code = 500
+            return {"message": "Internal Server Error"}
+
+        for version, data in result["releases"].items():
+            for elm in data:
+                # Only wheel are supported
+                if elm["url"] not in self.download_url_mocked or not elm[
+                    "url"
+                ].endswith(".whl"):
```
grouigrokon, line 238: I think that `.tar.gz` and `.zip` are valid packages too.

tests/tests_e3/python/main_test.py
```diff
@@ -149,3 +150,76 @@ def test_star_requirements():
         pypi.add_wheel(wheel3.path)
         pypi.add_requirement("src2==1.0.0")
         assert len(pypi.closure()) == 2
+
+
+def test_yanked(pypi_server):
```
grouigrokon, line 155: Maybe tests could be simpler using [pytest.mark.parametrize](https://docs.pytest.org/en/7.1.x/reference/reference.html#id23)

tests/tests_e3/conftest.py
leocardao: You should be aware that all exceptions are caught by requests-mock and transformed into "Internal Server Error". In my tests, I don't seem to have seen the error message added, only the standard http message for this error.

Having said that, at least in terms of code readability, I think you're right. So I'm going to change all that.

tests/tests_e3/conftest.py
leocardao: For the record: pypi returns a json `{ "message": "Not Found" }` if it can't find the requested project. That said, following this behavior for testing is not necessarily ideal, and it's unlikely that we'll ever need to test pypi's behavior in the event of an error.

tests/tests_e3/conftest.py
leocardao: For pypi, yes, these are obviously valid packages, but not for `e3-pypi-closure`, which only takes wheels into account.

Currently, the pypi simulator is only implemented to test `e3-pypi-closure`.

If I manage all file types, we'll have code that will never be executed, and therefore potentially completely buggy.

I think it's better to leave the implementation of these cases to the dev who will need it one day.

tests/tests_e3/python/main_test.py
leocardao: I don't understand what's in it for us (or even how to use it in this case).

Can you explain what you have in mind?

grouigrokon: I think the `requests-mock` package should be added to `setup.py`'s test dependencies too.
```python
@pytest.mark.parametrize(
    "arguments,expected",
    [
        ((None, None), "setuptools_scm-7.1.0-py3-none-any.whl"),
        ((["setuptools-scm"], None), "setuptools_scm-8.0.0-py3-none-any.whl"),
        ((["setuptools_scm"], None), "setuptools_scm-8.0.0-py3-none-any.whl"),
        ((None, "setuptools_scm==8"), None)
    ]
)
def test_yanked(pypi_server, arguments, expected):
    allowed_yanked, invalid_wheel = arguments
    expected_wheel = expected

    with pypi_server:
        mkdir("cache")

        with PyPIClosure(
            python3_version=11,
            platforms=[
                "x86_64-linux",
            ],
            cache_dir="cache",
            cache_file="pypi.cache",
            allowed_yanked=allowed_yanked,
        ) as pypi:
            if invalid_wheel:
                with pytest.raises(
                    PyPIError,
                    match=(
                        "Cannot find latest version for 'setuptools-scm': "
                        "No more suitable version"
                    ),
                ):
                    pypi.add_requirement(invalid_wheel)
            else:
                pypi.add_requirement("setuptools_scm >= 6.2, <= 8")
                all_filenames = [basename(f) for f in pypi.file_closure()]
                assert expected in all_filenames
```

tests/tests_e3/conftest.py
```diff
@@ -164,3 +176,109 @@ def pytest_sessionfinish(session, exitstatus):
...
+
+        try:
+            with open(path) as json_file:
+                result = json_loads(json_file.read())
+
+            if "releases" not in result:
+                raise Exception("Bad json metadata: 'releases' key not found")
+        except Exception as e:
+            context.status_code = 500
+            return {
```
grouigrokon, line 248: Returning a dict in a method which should return an `str` :)

tests/tests_e3/python/main_test.py
```diff
@@ -149,3 +150,46 @@ def test_star_requirements():
...
+        ((["setuptools-scm"], None), "setuptools_scm-8.0.0-py3-none-any.whl"),
+        ((["setuptools_scm"], None), "setuptools_scm-8.0.0-py3-none-any.whl"),
+        ((None, "setuptools_scm==8"), None),
+    ],
+)
+def test_yanked(pypi_server, arguments, expected):
+    allowed_yanked, invalid_wheel = arguments
+    expected_wheel = expected
+
+    rm("cache", recursive=True)
```
grouigrokon, line 168: Are you sure we have to remove them ? I mean that it should be a new tests started, thus its own temp folder.

tests/tests_e3/conftest.py
leocardao: Common error when adding types to a file ignored by mypy :')

tests/tests_e3/python/main_test.py
leocardao: I'm not sure, I just thought that when in doubt it would be more robust and avoid mistakes. The cache is used to avoid operations that I want to test.

grouigrokon: I confirm that each test (aka for each set of parameters) is run in its own tmp directory. You may want to test :)

grouigrokon (approved):

tox.ini
```diff
@@ -7,6 +7,7 @@ deps =
      ptyprocess;platform_system == "Linux"
      cov: pytest-cov
      cov: coverage
+     requests-mock
```
grouigrokon, line 10: This should also be added to the `test` dependencies in `setup,py`

grouigrokon (approved):

PR #675, opened 2024-01-24
Sort elements in SPDX document

grouigrokon:
In order to ease reading a diff between two versions of an SPDX document, both the tag:value and JSON formats of the document are now sorted.


src/e3/spdx.py
```diff
@@ -957,12 +962,41 @@ def add_section(section: str) -> None:
...
+                pkg_ids.append(rel.related_spdx_element)
+            else:
+                relas.append(rel.to_tagvalue())
+
+        output += relas
+
+        packages: list[str] = ["", ""]
+
+        for pkg in sorted(self.packages.values(), key=lambda package: package.name):
+            print(f"Package ID: {pkg.spdx_id}, Document ID: {self.spdx_id}")
```
enzbang, line 990: Is this print statement expected?

src/e3/spdx.py
```diff
@@ -957,12 +962,41 @@ def add_section(section: str) -> None:
...
+            if pkg.spdx_id in pkg_ids:
+                print("Adding main package")
```
enzbang, line 992: Same here

src/e3/spdx.py
grouigrokon: :'( In the middle of the stage :'( Sorry ...

src/e3/spdx.py
grouigrokon: Yes :) Why did I miss it again ... my bad :'(

src/e3/spdx.py
grouigrokon: Sorry, due to the presence of the `platform` package in `e3`, it became impossible to debug `e3` with PyCharm. So I added logs.\
I have to dig the configuration to make it possible (again) to debug `e3` ...

enzbang (approved):

PR #676, opened 2024-02-02
Fix pre-release version management

leocardao:
packaging.specifiers.Specifier ignore pre-release if the 'prereleases' parameter is set to None or False.

This meant that in some situations, a pre-release package authorised with the --allow-prerelease option was ignored even if the version is mandatory.

### DIFF
--- a/src/e3/python/pypi.py
+++ b/src/e3/python/pypi.py
@@ -7,7 +7,7 @@ import os
 import time
 import packaging.version
 import packaging.tags
-import packaging.specifiers
+from packaging.specifiers import Specifier
 from typing import TYPE_CHECKING
 from e3.error import E3Error
 from e3.python.wheel import Wheel
@@ -198,7 +198,7 @@ class Package:
                 v = packaging.version.parse(version)
 
                 if (not v.is_prerelease and not v.is_devrelease) or (
-                    v.is_prerelease and self.name in pypi.allowed_prerelease
+                    v.is_prerelease and self.name in self.pypi.allowed_prerelease
                 ):
                     self.versions.append(v)
             except Exception:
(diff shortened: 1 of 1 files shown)


src/e3/python/pypi.py
```diff
@@ -198,7 +198,7 @@ def __init__(
                 v = packaging.version.parse(version)
 
                 if (not v.is_prerelease and not v.is_devrelease) or (
-                    v.is_prerelease and self.name in pypi.allowed_prerelease
+                    v.is_prerelease and self.name in self.pypi.allowed_prerelease
```
grouigrokon, line 201: Knowing `self.pypi` is set to `pypi`, this change is not mandatory.

src/e3/python/pypi.py
```diff
@@ -357,9 +357,10 @@ def add_constraint(self, requirement: Requirement) -> None:
             self.versions = [
                 v
                 for v in self.versions
-                if packaging.specifiers.Specifier(f"{spec[0]}{spec[1]}").contains(
-                    str(v)
-                )
+                if packaging.specifiers.Specifier(
```
grouigrokon, line 360: Knowing `packaging.specifiers` is used only here, we should rather import through `from packaging.specifiers import Specifier`, and use `Specifier` here.

grouigrokon (dismissed):
My comments are only comments.
Feel free to resolve them directly.
Other than that, it seems ok.

grouigrokon:
Another remark (but this one I would like it to be fixed), is that the merge request description/commit message are really too short from my point of  view.

src/e3/python/pypi.py
leocardao: I agree, I will make the change.

leocardao: Done.

leocardao:
> This meant that in some situations, a pre-release package authorised with the --allow-prerelease option was ignored even if the version is mandatory.

@grouigrokon I don't mind putting in more information, but to be honest I don't really know what more to say ... I can write more precisely the problem it causes, but it will be totally related to the internal use of the tool, which wouldn't make much sense to explain in a public repo.

grouigrokon (approved):

PR #679, opened 2024-02-13
Add JsonData class to json module

grouigrokon:
A JSON object helper to use when handling some JSON data object.

Tests have been added to fully cover the class, and the module.

### DIFF
--- a/src/e3/json.py
+++ b/src/e3/json.py
@@ -6,14 +6,86 @@ import os
 
 import e3.error
 
+from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any
+    from typing import Any, TypeVar
+
+    JsonDataSelf = TypeVar("JsonDataSelf", bound="JsonData")
 
 
 class JsonError(e3.error.E3Error):
     pass
+
+
+class JsonDataInvalidJsonError(e3.error.E3Error):
+    """An error thrown when input data string does not represent a dictionary."""
+
+    pass
+
+
+class JsonData(ABC):
+    """An object to represent JSON data content."""
+
+    @abstractmethod
+    def as_dict(self) -> dict[str, object]:
+        """Return the dict representation of this JSON data object."""
+        ...
+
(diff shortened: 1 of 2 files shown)


src/e3/json.py
```diff
@@ -6,16 +6,71 @@
...
+        """Return the dict representation of this JSON data object."""
+        ...
+
+    def __eq__(self, other: object) -> bool:
+        """Check if this JSON data is identical to *other*.
+
+        :param other: The object to compare this JSON data with.
+
+        :return: A :class:`python:bool` set to **True** if both JSON data are identical,
+            **False** if they are not, or if *other* is not a CPE.
```
grouigrokon, line 36: Arf ... who wrote that !

src/e3/json.py
```diff
@@ -6,16 +6,71 @@
...
+        """Load a JSON string as a JSON data object.
+
+        :param content: The JSON string to initialize the JSON data object with.
+
+        :return: A new :class:`JsonData` object initialized with
+            values from the input dictionary.
+
+        .. seealso:: :meth:`from_dict`
+        """  # noqa RST304
+        return cls.from_dict(json.loads(content))
```
pierretr, line 72: A JSON string does not always represent a dict, it can be a list or even a simple string.
If this class is meant to work only with dicts maybe  the "dict" constraint should appear
in the class name and/or method names (`from_json` -> `from_json_dict` for example).

grouigrokon: I'd rather update the doc string to state that _content must be loadable through a call to `json.loads()`_

enzbang:
``` 
>>> json.loads('["a_json"]')
['a_json']
```

grouigrokon: In any case, this is more a doc issue. Naming it `from_json_dict()` will make (at least me) think it loads a `dict`, not a string.

pierretr: In addition to doc maybe also raise a specific exception when `json.loads` does not return a dict with a clear message indicating that only dictionaries are supported.

pierretr (approved):

PR #681, opened 2024-02-14
Implementation of VEX specifications version 1.0.0

grouigrokon:
# VEX specifications

The VEX (Vulnerabilities Exploitability Exchange) specifications have been added in the (new) `encoding` e3 module.

These specifications should help tracking the impact of CVEs on products, tacking for impact, justifications or actions to take for a given vulnerability.

# Doc warning fixes

Seems like `sphinx` imports the documentation of the `python:enum.Enum` class for classes which implement it, but do not have a documentation string.

Unfortunately, the base documentation has a warning ...
    
By simply adding a documentation string to two classes, the warnings got fixed.

### DIFF
--- a/src/e3/encoding/vex.py
+++ b/src/e3/encoding/vex.py
@@ -1,0 +1,990 @@
+"""Vulnerabilities Exploitability Exchange specifications.
+
+This specification package is based on
+https://www.cisa.gov/sites/default/files/2023-04/minimum-requirements-for-vex-508c.pdf.
+
+The document
+https://www.cisa.gov/sites/default/files/2023-01/VEX_Use_Cases_Aprill2022.pdf
+has also been used for a better understanding of some VEX implementations.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from dateutil.parser import parse as date_parse
+from enum import Enum
+from packaging.version import Version
+from typing import TYPE_CHECKING
+from uuid import uuid4
+
+from e3.json import JsonData
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+class ActionOrImpact(JsonData):
+    """An object to represent an action or impact for a statement status.
+
+    For *status* :attr:`Status.NOT_AFFECTED`, if *justification* is not
+    provided, then a VEX statement **MUST** provide an *impact_statement* that
+    further explains how or why the listed *product_id* s are *not affected* by
+    *vul_id*.
+
+    If *justification* is provided, then a VEX statement **MAY** provide an
+    *impact_statement*.
+
+    :ivar str | None statement: The statement of this action or impact.
+    :ivar datetime | None timestamp: The time at which the statement has
+        been last modified.
+    """  # noqa RST304
+
+    def __init__(
+        self,
+        statement: str | None = None,
+        timestamp: datetime | None = None,
+    ):
+        """Initialize an action or impact object.
+
+        :param str | None statement: The action or impact statement. If
+            **None**, it means that this object has no impact (or no action).
+        :param datetime | None timestamp: The time this action or impact has
+            been defined. If **None** and *statement* is defined, the timestamp
+            is set to the current time.
+        """
+        self.timestamp: datetime | None
+        self.statement: str | None = statement
+        if self.statement and not timestamp:
+            self.timestamp = datetime.now(timezone.utc).replace(microsecond=0)
+        elif isinstance(timestamp, datetime):
+            self.timestamp = timestamp.replace(microsecond=0)
+        else:
+            # Should be None
+            self.timestamp = None
+
+    def __bool__(self) -> bool:
+        """Check if this action or impact is defined.
+
+        An action or impact without a statement is considered as False.
+
+        :return: **True** if this action or impact has a statement defined to
+            something else that **None**, **False** else.
+        """  # noqa RST304
+        return self.statement is not None
+
+    def as_dict(self) -> dict[str, Any]:
+        return {
+            "statement": self.statement,
+            "timestamp": None
+            if self.timestamp is None
+            else self.timestamp.strftime("%Y-%m-%dT%H:%M:%SZ"),
+        }
+
+    @classmethod
+    def from_dict(cls, obj: dict) -> ActionOrImpact:
+        return cls(
+            statement=obj["statement"],
+            timestamp=date_parse(obj["timestamp"]) if obj["timestamp"] else None,
+        )
+
+
+class AuthorRole(Enum):
+    """Role of the document author.
+
+    The author role **MAY** specify the role of the *author*.
+
+    The author role **MAY** use the *category of publisher* roles defined by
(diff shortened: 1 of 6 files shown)


src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+        :raise: :exc:`python:ValueError` If *value* is not one of the possible
+            values of this enumerate, or if *default* has an invalid value and
+            *value* is **None**.
+        """  # noqa RST304
+        if isinstance(value, str) and value:
+            return cls(value)
+        elif isinstance(value, cls):
+            return value
+        elif default:
+            # The default value is either a Product status or a string, better
```
grouigrokon, line 278: Damned ... hide this copy/paste I should not see !

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+from enum import Enum
+from typing import TYPE_CHECKING, TypeVar
+from uuid import uuid4
+
+from e3.json import JsonData
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    DocumentSelf = TypeVar("DocumentSelf", bound="Document")
```
Nikokrock, line 24: I don't think this is needed. No need to put type on cls argument in classmethod

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+
+    :ivar str | None statement: The statement of this action or impact.
+    :ivar datetime | str | None timestamp: The time at which the statement has
+        been last modified.
+    """  # noqa RST304
+
+    def __init__(
+        self,
+        statement: str | None = None,
+        timestamp: datetime | str | None = None,
```
Nikokrock, line 51: Does the API needs to support string everywhere ? Can't we just have datetime | None ? (applies in several places)

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+        else:
+            # Should be None
+            self.timestamp = None
+
+    def __bool__(self) -> bool:
+        """Check if this action or impact is defined.
+
+        An action or impact without a statement is considered as False.
+
+        :return: A :class:`python:bool` set to **True** if this action or impact
```
Nikokrock, line 70: Description is a bit too convoluated. Just say: return True if ... else return False. It's clear for everyone

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+            has a statement defined to something else that **None**, **False**
+            else.
+        """  # noqa RST304
+        return self.statement is not None
+
+    def as_dict(self) -> dict[str, object]:
```
Nikokrock, line 76: I think the signature should be dict[str, Any]

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+            for st in statements:
+                self.add_statement(st)
+
+    def as_dict(self) -> dict[str, object]:
+        return {
+            "metadata": self.metadata.as_dict(),
+            "statements": [st.as_dict() for st in self.statements],
+        }
+
+    def add_statement(self, new_statement: Statement) -> None:
```
Nikokrock, line 171: missing docstring

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+        product.
+    :ivar list[Statement] | None statements: The list of statements defined for
+        this VEX document.
+    """
+
+    def __init__(
+        self,
+        metadata: Metadata,
+        statements: list[Statement] | None = None,
+    ):
```
Nikokrock, line 157: missing docstring

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+        information. Contrast with *author*.
+    :ivar datetime created_on: The time this VEX document was created at.
+    :ivar datetime last_updated_on: The time this VEX document was updated.
+    :ivar int version: The version of a VEX document.
+    :ivar str spec_version: The VEX requirements version used by the parent
+        document. By default, it is set to :attr:`SPEC_VERSION`.
+    """  # noqa RST304
+
+    AUTHOR: str = "AdaCore"
+    """Default document author."""
```
Nikokrock, line 321: Don't use triple quote here and for documentation of internal constant

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+    AUTHOR_ROLE: AuthorRole = AuthorRole.VENDOR
+    """Default document author role."""
+    SPEC_VERSION: str = "1.0.0"
+    """The specification version followed by this implementation of the VEX
+    specification.
+    """
+    VERSION: int = 1
+    """Default document version."""
+
+    class SpecVersion(object):
```
Nikokrock, line 331: Can't we reuse Version object from Python here ?

src/e3/encoding/vex.py
```diff
@@ -0,0 +1,902 @@
...
+    def __init__(
+        self,
+        author: str = AUTHOR,
+        author_role: AuthorRole | str = AUTHOR_ROLE,
+        tooling: str | None = None,
+        version: int = VERSION,
+        _id: str | None = None,
+        created_on: datetime | str | None = None,
+        last_updated_on: datetime | str | None = None,
+        spec_version: str | None = SPEC_VERSION,
```
Nikokrock, line 383: Missing docstring

src/e3/encoding/vex.py
grouigrokon: Done. All `datetime | str` have been replaced with simple `datetime`. All `from_dict()` had to be created accordingly.

src/e3/encoding/vex.py
grouigrokon: Done, removed all the `Self` definitions.

src/e3/encoding/vex.py
grouigrokon: I cant remember where I saw that we should now be using `object` rather than `Any`. As I cannot find it, I will use `Any` then.

grouigrokon: Done

src/e3/encoding/vex.py
grouigrokon: After adding all the docstrings to `__init__()` methods, I saw that the doc build does not include them :)

src/e3/encoding/vex.py
grouigrokon: Done

src/e3/encoding/vex.py
grouigrokon: Done

src/e3/encoding/vex.py
grouigrokon: Done

src/e3/encoding/vex.py
grouigrokon: Done

PR #682, opened 2024-02-26
e3.cve: Close NVD cache session

leocardao:
When it's not done, this can cause errors on Windows. Windows does not agree to delete a file or directory that is currently open.

### DIFF
--- a/src/e3/cve.py
+++ b/src/e3/cve.py
@@ -64,6 +64,7 @@ class NVD:
                 "the use of an API key for the NVD API is strongly recommended"
                 " to avoid rate limits"
             )
+        self._session: Session | None = None
 
     def search_by_cpe_name(
         self,
@@ -110,17 +111,19 @@ class NVD:
             else:
                 break
 
-    @cached_property
-    def session(self) -> Session:
+    def __enter__(self) -> Any:
         """Return an http requests Session supporting cache.
 
         Use requests_cache CachedSession when cache is requested.
         """
+        if self._session is not None:
+            return self
+
         if self.cache_db_path:
             from requests_cache import CachedSession
             from datetime import timedelta
 
-            session = CachedSession(
+            self._session = CachedSession(
                 self.cache_db_path,
                 backend=self.cache_backend,
                 # Use Cache-Control headers for expiration, if available
@@ -133,6 +136,27 @@ class NVD:
                 match_header=False,
             )
             logger.debug(f"using requests cache from {self.cache_db_path}")
-            return session
         else:
-            return Session()
+            self._session = Session()
+        return self
+
+    def __exit__(self, _type: Any, _value: Any, _tb: Any) -> None:
+        self.close()
+
+    def close(self) -> None:
+        if self._session is not None:
+            self._session.close()
+            self._session = None
+
+    @property
+    def session(self) -> Session:
+        if self._session is None:
+            from warnings import warn
+
+            warn(
+                "Using NVD.session without using `with` statement is deprecated",
+                DeprecationWarning,
+            )
+            self.__enter__()
+
+        return self._session  # type: ignore[return-value]
--- a/tests/tests_e3/cve/cve_test.py
+++ b/tests/tests_e3/cve/cve_test.py
@@ -6,21 +6,41 @@ import os
 def test_nvd_cve_search(socket_disabled):
     """Test the CVE DB research using cached data."""
     from requests_cache import NEVER_EXPIRE
+    from warnings import catch_warnings, simplefilter as warn_filter
 
     cache_db = os.path.join(os.path.dirname(__file__), "cache")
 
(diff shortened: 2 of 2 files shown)


grouigrokon:
I do not see the rationale behind that.
Any hint on why we should do that ?
Your changes simply lose the point of caching the session.

src/e3/cve.py
```diff
@@ -79,6 +80,9 @@ def search_by_cpe_name(
         :param results_per_page: number of results to return for each request,
              note that it is recommended to keep the default setting
         """
+        if self.session is None:
+            raise ValueError("NVD session not started")
```
grouigrokon, line 84: What about: *NVD session not Stated: the NVD class should be used in a `with` statement* (or something like that)

leocardao: Done.

grouigrokon:
I know it makes a bit more work, and more reviews ;) but: I would keep the old behavior (with the potential bug) still possible. With these modifications, there is a backward compatibility issue.

What is your advice with that ?

leocardao:
> I know it makes a bit more work, and more reviews ;) but: I would keep the old behavior (with the potential bug) still possible. With these modifications, there is a backward compatibility issue.
> 
> What is your advice with that ?

I don't really have any advice at this level, it's pretty trivial to do. In the end, I'd be in favor of adding a method close to ensure consistency in the class and a warning to notify the dev.

In fact, it's always possible to get the bug by calling the `__enter__` method explicitly (without using the `with` keyword) without calling the `__exit__` method. But I don't think that's what you're looking for :)

I suggest implementing the session property, which will open the session if it hasn't already been done, but with a `DeprecationWarning` if it has, to notify the dev that this is not desired (and that one day we may remove this feature). For me, this is the best way to go.

Then we can always use `log.warnings`, but from my point of view, we shouldn't use logs for that.

grouigrokon:
> I don't really have any advice at this level, it's pretty trivial to do. In the end, I'd be in favor of adding a method close to ensure consistency in the class and a warning to notify the dev.
> 
> In fact, it's always possible to get the bug by calling the `__enter__` method explicitly (without using the `with` keyword) without calling the `__exit__` method. But I don't think that's what you're looking for :)
> 
> I suggest implementing the session property, which will open the session if it hasn't already been done, but with a `DeprecationWarning` if it has, to notify the dev that this is not desired (and that one day we may remove this feature). For me, this is the best way to go.
> 
> Then we can always use `log.warning`, but from my point of view, we shouldn't use logs for that.
> 
> Does that make sense to you, @grouigrokon?

That would be ok to me, and I would use warnings ;) There are warnings for the use of the NVD API key ... so why not on that too ?

src/e3/cve.py
```diff
@@ -94,7 +101,7 @@ def search_by_cpe_name(
 
         start_index = 0
         while True:
-            r = self.session.get(url + f"&startIndex={start_index}", headers=headers)
+            r = self._session.get(url + f"&startIndex={start_index}", headers=headers)
```
grouigrokon, line 104: To me we should call for `self.session` here.

leocardao: I didn't say anything, it doesn't make sense, or would force me to have the same warning twice :)

src/e3/cve.py
```diff
@@ -79,6 +80,12 @@ def search_by_cpe_name(
         :param results_per_page: number of results to return for each request,
              note that it is recommended to keep the default setting
         """
+        if self._session is None:
```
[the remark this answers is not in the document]
leocardao: Done.

tests/tests_e3/cve/cve_test.py
```diff
@@ -6,21 +6,28 @@
...
+            nvd_db.session.cache.reset_expiration(NEVER_EXPIRE)
+            cve_urls = [
+                cve.nvd_url
+                for cve in nvd_db.search_by_cpe_name(
+                    "cpe:2.3:a:libpng:libpng:1.6.0:-:*:*:*:*:*:*", results_per_page=5
+                )
+            ]
+
+        assert len(w) == 0, "An unexpected python warning detected"
+
```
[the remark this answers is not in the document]
leocardao: Done.

tests/tests_e3/cve/cve_test.py
```diff
@@ -6,21 +6,41 @@
...
+        _ = tmp.session
+
+        assert tmp._session is not None, "tmp._session is None"
+
+        assert len(w) == 1, f"No python warning occure: len(warnings) == {len(w)}"
+        assert issubclass(w[-1].category, DeprecationWarning)
+
+        tmp.close()
+
+        assert tmp._session is None, "tmp._session is not None"
```
grouigrokon, line 42: Some IDEs may warn that you are using internal variables ... But that's ok. Not sure `mypy` has a warning for that :)

PR #689, opened 2024-03-11
Fix typing information for Checkout events

yakobowski:
The possible values in the dictionary are not necessarily strings: the release_branching_info field is actually a dict.

### DIFF
--- a/src/e3/anod/action/__init__.py
+++ b/src/e3/anod/action/__init__.py
@@ -187,14 +187,17 @@ class Checkout(Action):
 
     __slots__ = ("uid", "repo_name", "repo_data")
 
-    def __init__(self, repo_name: str, repo_data: dict[str, str]):
+    def __init__(self, repo_name: str, repo_data: dict[str, Any]):
         """Initialize a Checkout object.
 
         :param repo_name: The name of the repository.
         :param repo_data: A dictionary with the following keys:
             - 'url': The repository URL;
             - 'revision': The revision to checkout;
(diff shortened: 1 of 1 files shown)


src/e3/anod/action/__init__.py
```diff
@@ -187,14 +187,17 @@ class Checkout(Action):
...
             - 'url': The repository URL;
             - 'revision': The revision to checkout;
-            - 'vcs': The Version Control System kind (a string).
+            - 'vcs': The Version Control System kind (a string);
+            - 'license': the license for the files contained in the repository;
+            - 'release_branching_info': a dictionary containing information
+              on how to branch the repository when a release is branched.
 
         At present, only 'git' is supported.
         """
```
leocardao, line 203: You obviously have an indentation problem... but I can't see it (it's pre-commit that says so).

leocardao (changes requested):

src/e3/anod/action/__init__.py
yakobowski: Indeed. Not sure why I convinced myself that my initial fix did work, but should now be ok.

yakobowski: There is a new error, but it seems unrelated to my change.

yakobowski: Also, I think the error reported is fixed by #68 @leocardao

leocardao (approved):

PR #690, opened 2024-03-18
Remove reference to obsolete features/packages

Nikokrock:


src/e3/decorator.py
```diff
@@ -51,72 +49,3 @@ def empty_func(*args: Any, **kargs: Any) -> None:
         del args, kargs
 
     return empty_func
-
-
-class memoize:
```
enzbang, line 56: We're not yet ready to drop memoize, still in use at AdaCore

Nikokrock: Only two references I think

Nikokrock: Almost remove but cannot wait in order to move to pyproject.toml. I have removed for the moment the commit

enzbang (approved):

PR #698, opened 2024-03-21
Enable Anod class to use methods instead of properties

leocardao:


src/e3/anod/spec.py
```diff
@@ -203,6 +203,34 @@ class MyProduct(Anod):
     ExternalSourceBuilder = e3.anod.package.ExternalSourceBuilder
     ThirdPartySourceBuilder = e3.anod.package.ThirdPartySourceBuilder
 
+    def __new__(cls, *args: Any, **kwargs: Any) -> Any:
+        """Replace `method` by property when decorator is missing."""
+        if callable(cls.enable_name_generator):
```
grouigrokon, line 208: This could be factorized.
```python
should_be_property = ("readme_info", "build_space_name", ...)
for prop ins should_be_property:
    class_property = getattr(cls, prop)
    if callable(class_property):
        setattr(cls, prop, property(fget=class_property)
```
(or something like that)

tests/tests_e3/anod/spec_test.py
```diff
@@ -147,3 +147,12 @@ class GeneratorDisabled(Anod):
...
     assert spec_disable.args == {"q1": ""}
+
+
+def test_missing_property():
+    class NoProperty(Anod):
+        def source_pkg_build(self) -> list:
+            return []
+
+    noproperty = NoProperty(qualifier="", kind="source")
+    assert noproperty.source_pkg_build == []
```
grouigrokon, line 158: We may want a bit more here. I know that I sometimes forgot the `@property` for `build_deps` (for instance), and calling the `anod.build()` failed. Would it be a good test ?

Nikokrock: we don't have a full anod currently in e3-core though

grouigrokon: Oh right ... sorry :)

src/e3/anod/spec.py
leocardao: Done.

grouigrokon (approved):

PR #703, opened 2024-03-26
Create a pytest plugin to reuse fixtures in other projects

enzbang:
Remove deprecated requirements traceability which is not maintained.

Create e3.pytest to reuse part of the code in conftest.py in other projects.
Document the new plugin.

Note that projects that rely on e3-core conftest.py files must be updated to use the plugin instead.


docs/source/pytest.rst
```diff
@@ -0,0 +1,85 @@
...
+
+env_protect
+^^^^^^^^^^^
+
+When activated, the plugin will register the ``env_protect`` feature to ensure
+that all tests are run in isolation. All changes to the environment done in
+each test won't impact other tests. Also, each test is run in a separate temp
+directory, you won't have to cleanup the files that the tests create.
+
+``env_protect`` also set some environment variables such as:
```
grouigrokon, line 28: Should be *also **sets***

docs/source/pytest.rst
```diff
@@ -0,0 +1,85 @@
...
+Specific test for the windows platform are also detected:
+
+* ``if sys.platform == win32``
+* ``if sys.platform != win32``
+* ``unix-only``
+
+You can also skip complete files by creating an ``omit file`` in
+``tests/coverage/omit-file-<os>``. The file should contain a filename per line.
+
+Finally, the option ``--e3-cov-rewrite <origin> <dest>`` change the paths
```
grouigrokon, line 64: Should be ***changes** the paths*

src/e3/pytest.py
```diff
@@ -0,0 +1,223 @@
...
+            if o != os_name
+        ):
+            cov.exclude(regex)
+
+        # Handling of Windows
+        if os_name == "windows":
+            cov.exclude('if sys.platform != "win32":')
+        else:
+            cov.exclude('if sys.platform == "win32":')
+            cov.exclude("unix: no cover")
```
grouigrokon, line 152: Works for Darwin too (I guess you tested it, but just to make sure ...)

src/e3/pytest.py
```diff
@@ -0,0 +1,235 @@
...
+        e.g. src/
+    :param cov_db: path to the .coverage data base
+    """
+    paths = PathAliases()
+    paths.add(origin_dir, new_dir)
+
+    old_cov_file = NamedTemporaryFile(dir=os.path.dirname(cov_db))
+    old_cov_file.close()
+    try:
+        mv(cov_db, old_cov_file.name)
```
grouigrokon, line 192: Is it in the right way ?
We move `cov_db` to `old_cov_file.name` and finally remove `old_cov_file.name` ?

src/e3/pytest.py
```diff
@@ -0,0 +1,235 @@
...
+
+        old_coverage_data = CoverageData(old_cov_file.name)
+        old_coverage_data.read()
+        new_coverage_data = CoverageData(cov_db)
+        new_coverage_data.update(old_coverage_data, aliases=paths)
+        new_coverage_data.write()
+    finally:
+        os.unlink(old_cov_file.name)
```
grouigrokon, line 200: I think the `old_cov_file.unlink()` method may be used.

src/e3/pytest.py
```diff
@@ -0,0 +1,235 @@
...
+    When the variable results_dir is set to an existing directory, the testsuite
+    will generate results file in "anod" format.
+    """
+    global test_errors
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    rep = outcome.get_result()
+    results_dir = os.environ.get("RESULTS_DIR")
+
+    if not results_dir or not os.path.isdir(results_dir):
```
grouigrokon, line 216: If the directory does not exist, I think we could either let the user know it, or create it.

grouigrokon (commented):
Please check the comments I made. Do not hesitate to reject them if you think they are just wrong.

src/e3/pytest.py
enzbang: I don't get your comment. If this either to activate coverage on windows only, or activate coverage on other platforms.

src/e3/pytest.py
enzbang: What is the right way? ;) 

This is just moving the coverage file in another location and then deleting the extra file. Note that this is not a new code.

src/e3/pytest.py
enzbang: AttributeError: '_io.BufferedRandom' object has no attribute 'unlink'

docs/source/pytest.rst
enzbang: Thank you. Fixed!

docs/source/pytest.rst
enzbang: Fixed!

src/e3/pytest.py
enzbang: Yes, macOS and Linux are considered unix here.

grouigrokon (approved):
Since all my comments have been resolved.

PR #712, opened 2024-04-04
Fix pytest error

leocardao:
When pytest is used with xdist and junitxml, tests are not run in the main thread. In addition, signals can only be used in the main thread.

This results in the failure of some tests that attempt to define a signal handler.

The signal handler causing the error is not tested, so we simply disable the handler if the current thread is not the main thread.

### DIFF
--- a/src/e3/main.py
+++ b/src/e3/main.py
@@ -31,6 +31,7 @@ import logging
 import os
 import signal
 import sys
+import threading
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


src/e3/main.py
```diff
@@ -144,7 +144,11 @@ def sigterm_handler(sig: int, frame: FrameType | None) -> NoReturn:  # unix-only
             raise SystemExit("SIGTERM received")
 
         if sys.platform != "win32":  # unix-only
-            signal.signal(signal.SIGTERM, sigterm_handler)
+            import threading
```
enzbang, line 147: Given that this will be executed by most users of e3, I don't think a huge added value to defer the import of the threading module. Why not moving it to the top of the file?

leocardao: Done.

grouigrokon (approved):

PR #714, opened 2024-04-04
E3 pypi closure

enzbang:
Silence "cannot parse version" warning by default
    
This message occurs way too often when running e3-pypi-closure
because of bad version in old software.


grouigrokon (approved):
Shall the `NEWS.md` file be updated ?

Issue #717, opened 2024-04-10
Iterate a DAG with cycles with the "enable_busy_state" loops


Jicquel:
A simple reproducer:

```
d = DAG()
d.add_vertex("a")
d.add_vertex("b")
d.update_vertex("a", predecessors=["b"])

it = DAGIterator(d, enable_busy_state=True)
[k for k, _ in it]
```

The iteration will never end, as the loop is not detected. Non-visited nodes are waiting for predecessors as if the latter were in the `busy` state.

Jicquel:
I did not paste the correct reproducer: here it is:

```
d = DAG()
d.add_vertex("a")
d.add_vertex("b")
d.update_vertex("a", predecessors=["b"])
# Force the creation of a cycle
d.update_vertex("b", data="newb", predecessors=["a"], enable_checks=False)

it = DAGIterator(d, enable_busy_state=True)
[k for k, _ in it]
```

As the comment explains (coming from tests/tests_e3/collection/dag/main_test.py), checks are implicit during vertices insertion and modification. I will close this issue, as the checks are made before iterating over the DAG.

PR #718, opened 2024-04-10
Detect cycles during DAG iteration when "enable_busy_state" is set

Jicquel:
When "enable_busy_state" is set, iterate over the DAG made all non-visited nodes wait for predecessors as if the latter were in the busy state.

Cycles are now always checked, no matter the enable_busy_state value.

Fix issue #717

### DIFF
--- a/src/e3/collection/dag.py
+++ b/src/e3/collection/dag.py
@@ -75,19 +75,18 @@ class DAGIterator:
         result = next((k for k in self.non_visited if self.pred_number[k] == 0), None)
 
         if result is None:
+            for node in self.non_visited:
+                minimal_cycle = self.dag.shortest_path(node, node)
+                if minimal_cycle is not None:
+                    raise DAGError(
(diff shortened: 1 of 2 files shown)


src/e3/collection/dag.py
```diff
@@ -75,19 +75,18 @@ def next_element(
         result = next((k for k in self.non_visited if self.pred_number[k] == 0), None)
 
         if result is None:
+            for node in self.non_visited:
```
Nikokrock, line 78: when busy_state is enabled it's legal to get a None (when no nodes are ready)

Jicquel: The git preview does not help, but the line 86 should ensure that if we do not have a cycle and we are in busy_state, then we can return None

Jicquel:
Checks are implicitly done during vertices insertion and modification. I will close this PR, as the checks are made before iterating over the DAG.

PR #719, opened 2024-04-18
Fix pip error

leocardao:
pip fails in production because it cannot find setuptools (which is installed) and tries to connect it to pypi.

We add the --no-build-isolation option to allow pip to use the setuptools installed in our python.

### DIFF
--- a/src/e3/python/wheel.py
+++ b/src/e3/python/wheel.py
@@ -58,6 +58,7 @@ class Wheel:
                     "wheel",
                     ".",
                     "-q",
+                    "--no-build-isolation",
                     "--no-deps",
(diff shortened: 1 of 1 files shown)


grouigrokon:
Seems like it break python 12 tests ?

leocardao:
> Seems like it break python 12 tests ?

@grouigrokon I've seen it, the problem is that without it, it seems to break in production too ... It may not be the right solution, but I don't have a better one to propose at the moment (besides, I don't quite understand the error reported by the CI). I'll draft this MR in the meantime.

Nikokrock:
Closing PR (agreed with leo)

PR #721, opened 2024-05-14
Fix pytest when source coverage is not used

### LINKED ISSUE #720: Pytest may not run with coverage
When trying to run a single test with the latest `pytest` driver, it ends up with:

```
INTERNALERROR>   File ".../lib/python3.11/site-packages/_pytest/config/__init__.py", line 1709, in getoption
INTERNALERROR>     raise ValueError(f"no option named {name!r}") from e
INTERNALERROR> ValueError: no option named 'cov_source'
```

grouigrokon:
The source coverage option may not be defined when running the tests manually. The option existence should be checked before trying to get it.

The `coverage` package dependency was also missing from the `pyproject.toml`.

Closes #720


pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
 ]
 
 dependencies = [
+  "coverage",
```
enzbang, line 24: It's better to avoid such dependencies as you don't want to pollute the venv in production, lambdas, ...

That being said, a comment in the toml file would have been appropriate. I should have done it.

src/e3/pytest.py
```diff
@@ -98,9 +98,14 @@ def restore_env() -> None:
 
 
 def pytest_configure(config: pytest.Config) -> None:
-    if config.getoption("e3") and config.getoption("cov_source"):
+    if (
+        hasattr(config.option, "e3")
```
enzbang, line 102: Not needed, this come with e3-core

enzbang (changes requested):

src/e3/pytest.py
```diff
@@ -12,19 +12,26 @@
...
-from coverage import Coverage
-
-
-import pytest
+try:
+    import pytest
+    from coverage.sqldata import CoverageData
+    from coverage.files import PathAliases
+    from coverage import Coverage
+except ImportError as ie:
```
enzbang, line 20: Add # defensive code for this line

enzbang (approved):

PR #725, opened 2024-05-29
Handle coverage module backward incompatibility

grouigrokon:
The `CoverageData.update()` method _aliases_ argument has been replaced by _map_path_ in coverage 7.5.3 release (see 
[nedbat/coveragepy@c45ebac](https://github.com/nedbat/coveragepy/commit/c45ebac1b83656bfdd92d6675f08fa33627785e6))

Catch the exception and call the method with old parameters if needed.

### DIFF
--- a/src/e3/pytest.py
+++ b/src/e3/pytest.py
@@ -209,7 +209,20 @@ def fix_coverage_paths(origin_dir: str, new_dir: str, cov_db: str) -> None:
         old_coverage_data = CoverageData(old_cov_file.name)
         old_coverage_data.read()
         new_coverage_data = CoverageData(cov_db)
-        new_coverage_data.update(old_coverage_data, aliases=paths)
+
+        # Before 7.5.3, the map_path argument (of type Callable[[str], str])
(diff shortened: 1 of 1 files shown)


src/e3/pytest.py
```diff
@@ -194,7 +194,16 @@ def fix_coverage_paths(origin_dir: str, new_dir: str, cov_db: str) -> None:
         old_coverage_data = CoverageData(old_cov_file.name)
         old_coverage_data.read()
         new_coverage_data = CoverageData(cov_db)
-        new_coverage_data.update(old_coverage_data, aliases=paths)
+        # Incompatibility with version 7.5.2 and 7.5.3 of the update() method.
```
pierretr, line 197:
```suggestion
        # aliases argument has been replaced by map_path in coverage 7.5.3 release.
```

enzbang: Can we do the opposite? Make the call work by default with the new version of coverage and have a fallback for legacy versions?

grouigrokon: Done

pierretr (approved):

PR #727, opened 2024-06-07
Fix e3-pypi-closure name comparison

leocardao:
Before this patch, if you had a requirement like 'Django==2.2', the version constraint would be ignored because the code was trying to compare 'Django' == 'django'

Now we don't have to worry about character case in this comparison, which corrects the problem.


Nikokrock:
isn't canonicalize_name enough ?

leocardao:
> isn't canonicalize_name enough ?

@Nikokrock Visibly not, because for django, we end up comparing ‘Django’ != ‘django’, which is true, and so we enter this ‘if’, and, as a result, it ignores the constraint.

liaiss (approved):

Nikokrock:
Call canonalize on self.name ?

leocardao:
> Call canonalize on self.name ?

Done :)

grouigrokon (approved):

PR #729, opened 2024-06-18
Add minor OS version to platform object

leocardao:


src/e3/env.py
```diff
@@ -267,7 +282,12 @@ def get_platform(
             elif split_value[0] == "target":
                 return saved_target
             elif not propagate_build_info:
-                return Platform.get(*split_value)  # type: ignore
+                return Platform.get(
+                    platform_name=split_value[0],
+                    version=split_value[1],
+                    machine=split_value[2],
```
grouigrokon, line 288: No minor version here ?

src/e3/env.py
```diff
@@ -279,7 +299,12 @@ def get_platform(
                     # Linux machine should not change the OS version
                     if split_value[1] is None:
                         split_value[1] = saved_build.os.version
-                return Platform.get(*split_value)  # type: ignore
+                return Platform.get(
+                    platform_name=split_value[0],
+                    version=split_value[1],
+                    machine=split_value[2],
```
grouigrokon, line 305: No minor version possible here ?

src/e3/os/platform.py
```diff
@@ -381,10 +383,31 @@ def get(
         # default (cross case)
         if version == UNKNOWN:
             if is_host:
+                SystemInfo.fetch_system_data()
                 version, kernel_version = SystemInfo.os_version()
+                if not version_minor:
```
grouigrokon, line 388: So we won't enter here if `version_minor` is `UNKNOWN`

src/e3/os/platform.py
```diff
@@ -381,10 +383,31 @@ def get(
...
+                    if "." in SystemInfo.ld_info["version"]:  # type: ignore[index]
+                        version_minor = SystemInfo.ld_info["version"].split(".")[1]  # type: ignore[index]
+                    elif "." in version:  # Can occure when host is windows
+                        version_minor = version.split(".")[1]
             else:
                 version = KNOWLEDGE_BASE.os_info[name]["version"]
+                if not version_minor and "." in version:
```
grouigrokon, line 395: Same here. Is that expected ?

src/e3/env.py
leocardao: Not in the current state at least, that would require another change (not necessarily very complex to have that said, but it wasn't in what was asked of me, and it would add a field to --target/--build/--host).

Note that my current change still allows you to get version_minor if the OS version in the parameter string uses a format like: MAJOR.MINOR[.MORE_SUB_VERSION]*

src/e3/os/platform.py
leocardao: Indeed, I will make the change.

leocardao: Done.

src/e3/os/platform.py
leocardao: Not expected. Fix done.

src/e3/env.py
leocardao: Same answer as for the next comment :)

Nikokrock:
Can we have a live conversation on the need for that change? The change seems a bit awkward in term of API.

src/e3/os/platform.py
```diff
@@ -167,15 +170,20 @@ def os_version(cls) -> tuple[str, str]:
             if "redhat" in name or "red hat" in name:  # os-specific
```
grouigrokon, line 170: Rather than repeating `ull_version_number = cls.ld_info["version"]`, what about setting it once and for all (before or after the `if` statement) like:
```python
distro_version: str = cls.ld_info["version"] if "version" in cls.ld_info else ""
```
I feel like the `version_number` could be set this way too.
I do not like the `full_version_number` name because the type of the variable seems to be a string.

leocardao:
Replaced by #736 due to some conflict

PR #730, opened 2024-06-19
Update NEWS.md for a new package version

grouigrokon:
This modification is to update the pypi package to a new version 22.5.1.


grouigrokon:
I do not have any message about that CLA re-signing stuff ...

PR #733, opened 2024-06-26
Make x86_64-windows64 the default platform for windows host

leocardao:


Nikokrock:
This patch is useless in practice as all computer will report x86_64 anyway

PR #734, opened 2024-06-26
Add a shared libraries closure check to anod spec

grouigrokon:
The `check_shared_libraries_closure()` methods allows to make sure the
shared library of a spec only link with valid shared libraries.

For instance, when linking with our own version of `zlib`, we want
to make sure that the built shared libraries do not link with
`/usr/lib/libz.so`, but with our `zlib`.

### DIFF
--- a/src/e3/anod/spec.py
+++ b/src/e3/anod/spec.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 import os
+import re
+import sys
+
 from packaging.version import Version
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 import yaml
@@ -14,6 +18,9 @@ import e3.os.process
 import e3.text
 from e3.anod.error import AnodError, ShellError
 from e3.anod.qualifiers_manager import QualifiersManager
+from e3.fs import find
+from e3.os.fs import which
+from e3.platform_db.knowledge_base import OS_INFO
 from e3.yaml import load_with_config
 
 # Default API version
@@ -25,7 +32,7 @@ SUPPORTED_API = (__version__, "1.5", "1.6")
 # Version 1.4 (initial version)
 # Version 1.5
 #    NEW: YAML files are searched in subdirectories whose basename is the
-#    the associated spec basename.
+#    associated spec basename.
 #    DEPRECATED: YAML files in same directory as the spec
 # Version 1.6
 #    NEW: Add support for spec function automatically declared inside each spec
@@ -62,7 +69,7 @@ if TYPE_CHECKING:
         BUILD_PRIMITIVE, DOWNLOAD_PRIMITIVE, INSTALL_PRIMITIVE, SOURCE_PRIMITIVE
     ]
 
-    # Supported primitivies are build, install, source, and test
+    # Supported primitives are build, install, source, and test
     PRIMITIVE = Union[
         BUILD_PRIMITIVE, INSTALL_PRIMITIVE, SOURCE_PRIMITIVE, TEST_PRIMITIVE
     ]
@@ -127,9 +134,9 @@ def fetch_attr(instance: Any, name: str, default_value: Any) -> Any:
     this does not hide AttributeError exceptions that getting an existing
     attribute might raise.
     """
-    # Singleton used to determine whether call to gettattr returns the default
-    # value (i.e. attribute is missing) or just returns the attribute value (it
-    # could be None).
+    # Singleton used to determine whether call to ``gettattr()`` returns the
+    # default value (i.e. attribute is missing) or just returns the attribute
+    # value (it could be ``None``).
     sentinel = object()
 
     return (
@@ -166,7 +173,7 @@ class Anod:
     :cvar sandbox: e3.anod.sandbox.SandBox object shared by all Anod instances
     :vartype sandbox: e3.anod.sandbox.SandBox | None
 
-    Some attributes are meant the be overwritten in the specification file:
+    Some attributes are meant to be overwritten in the specification file:
 
     :cvar source_pkg_build: a dictionary associating Anod.SourceBuilder to the
         Anod.Source names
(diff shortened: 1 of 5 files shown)


src/e3/anod/spec.py
```diff
@@ -387,6 +394,94 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
...
+
+        Make sure that the libraries only depend on authorized system shared
+        libraries.
+
+        Raise an error when a problem is detected.
+
+        :param prefix: The path where to find the library to be checked.
+        :param ignored_libs: The list of additional system libraries authorized
+            in the closure.
+        :param ldd_output: An ``ldd`` command output. Dict key is the name/path
```
grouigrokon, line 413: I guess I need to fix that documentation :)

src/e3/anod/spec.py
```diff
@@ -387,6 +394,101 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
...
+            *ldd_output*) is not in the same directory as the analysed element.
+        """
+        ignored: list[str] = ignored_libs or []
+        errors: dict[str, list[str]] = {}
+        lib_file: str = ""
+
+        if ldd_output is None:
+            # If ldd is not in the path, return and issue a warning.
+            if not which("ldd"):
+                self.log.warning("DLL closure check skipped: ldd could not be found.")
```
enzbang, line 434: Why not return an error in that case? The user might think everything is okay even though no check has been run.

src/e3/anod/spec.py
```diff
@@ -387,6 +394,101 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
             name=self.build_space_name, platform=self.env.platform
         )
 
+    def check_dll_closure(
```
enzbang, line 397: dll is windows specific, why not check_shared_library_closure

src/e3/anod/spec.py
```diff
@@ -387,6 +394,101 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
...
+            lib_files = re.findall(r"^([^\t].*):$", ldd_output, flags=re.M)
+
+        if self.sandbox and hasattr(self.sandbox, "root_dir"):
+            root_dir = self.sandbox.root_dir
+        else:
+            root_dir = str(Path(self.build_space.install_dir).parent)
+
+        # Retrieve list of shared lib dependencies for sanity checking
+
+        self.log.info("Running DLL closure check")
```
enzbang, line 455: shared libraries instead of DLL

src/e3/anod/spec.py
```diff
@@ -387,6 +394,101 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
...
+                # ldd for ``mydll.so``.
+                lib_file = line[:-1]
+            elif lib_file and " => " in line:
+                # Line looks like:
+                # ``        otherdll.so => /other/otherdll.so (0xabcd)``
+                name, path = line.strip().split(" => ", 1)
+                path = re.sub(" (.*)", "", path)
+                # Check if we should ignore it. On Windows, the paths are
+                # case-insensitive.
+                if sys.platform != "win32":
```
enzbang, line 470: case sensitive is file system dependent, not OS dependent

I would add a parameter with a good default value

src/e3/anod/spec.py
```diff
@@ -387,6 +394,101 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
...
+                    if lib_file not in errors:
+                        errors[lib_file] = []
+                    errors[lib_file].append(f"\n\t- {name}: {path}")
+
+        if errors:
+            # Build up error message
+            err_msg: str = "Dependencies on external libraries:"
+            for shared_lib, dependencies in errors.items():
+                err_msg = (
+                    f"{err_msg}\n    Share lib {shared_lib} has external shared "
```
enzbang, line 487: typo: Shared lib

src/e3/anod/spec.py
```diff
@@ -387,6 +394,94 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
...
+    def check_dll_closure(
+        self,
+        prefix: str | None = None,
+        ignored_libs: list[str] | None = None,
+        ldd_output: str | None = None,
+    ) -> None:
+        """Sanity check the shared library closure.
+
+        Make sure that the libraries only depend on authorized system shared
+        libraries.
```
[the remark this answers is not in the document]
grouigrokon: Well, we do have `ldd` on our Windows hosts. So I just think that a do update stating that `ldd` executable should be in the path for the closure to work.

src/e3/anod/spec.py
grouigrokon: Because it could be run on Windows which may not have `ldd` in the path, and (I think) we do not want the build to fail in that case.

src/e3/anod/spec.py
grouigrokon: Will do

src/e3/anod/spec.py
grouigrokon: Will do

src/e3/anod/spec.py
grouigrokon: Good point :) Used to associate case insensitive file system with windows ... My bad.

src/e3/anod/spec.py
enzbang: I think we want the build to fail if we're in prod and we're relying on that feature to ensure that the libs are correct. We could raise a warning in USER mode if the call to that function raise an error but keep the error in PROD mode. This function should raise an error if ldd is not found.

src/e3/anod/spec.py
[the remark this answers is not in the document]
enzbang: if we want to rely on Cygwin's dll then we should document it. And we should document that the function will only run on host having dll installed (so Linux and Windows with Cygwin - but not macOS for instance)

src/e3/anod/spec.py
grouigrokon: Done

src/e3/anod/spec.py
grouigrokon: Added the `case_sensitive` parameter.

src/e3/anod/spec.py
grouigrokon: Done

src/e3/anod/spec.py
```diff
@@ -387,6 +393,118 @@ def bind_to_sandbox(self, sandbox: SandBox) -> None:
...
+                # Line is like ``/mypath/mydll.so:``, it's the result of
+                # ldd for ``mydll.so``.
+                lib_file = line[:-1]
+            elif lib_file and " => " in line:
+                # Line looks like:
+                # ``        otherdll.so => /other/otherdll.so (0xabcd)``
+                name, path = line.strip().split(" => ", 1)
+                path = re.sub(" (.*)", "", path)
+                # Check if we should ignore it. On Windows, the paths are
+                # case-insensitive.
```
enzbang, line 485: instead of having that comment, it think it would be better to set the default value for case_sensitive to (sys.platform == "win32")

grouigrokon: I had to move the platform check in the method body, or the doc would not build.

tests/tests_e3/anod/spec_test.py
```diff
@@ -50,6 +153,35 @@ def build(self):
...
+    test_spec: Anod = Anod("", kind="install")
+    test_spec.sandbox = SandBox(root_dir=os.getcwd())
+    if ldd_output is None:
+        # Use the current executable lib directory.
+        exe_path: Path = Path(sys.executable)
+        lib_path: Path = Path(
+            exe_path.parent.parent, "lib" if sys.platform != "win32" else ""
+        )
+        test_spec.check_shared_libraries_closure(
+            prefix=str(lib_path), ignored_libs=ignored, ldd_output=ldd_output
```
enzbang, line 170: here ldd_output is None, so it would be better to write None for clarity

this test will fail on Windows system without ldd, it will also fail on macOS, can had a @pytest.skipif(not ldd)

grouigrokon: Do you prefer that I use `skipif` or that I define an `ldd = require_tool("ldd")` in `conftest.py` ? I would vote for the second, but to me both are equal.

enzbang: require_tool is fine with me

enzbang (approved):

PR #735, opened 2024-06-28
Use at least --shallow-since when git_shallow_fetch is enabled

pierretr:
When git_shallow_fetch feature is requested, even when we need to compute a changelog, we should not fetch the full depth of the repository. Just fetch at the required depth instead.


pierretr:
`--shallow-since` only works with dates.

